PR #2, opened 2014-05-17
Added initial Node is published condition.

paranojik:


src/Plugin/Condition/NodeIsPublished.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Provides a 'Node Is Published' condition.
+ *
+ * @Condition(
+ *   id = "rules_node_is_published",
+ *   label = @Translation("Node is published"),
+ *   context = {
+ *     "node" = {
+ *       "type" = "entity",
```
fago, line 20: This can be "entity:node" instead without the whole constraints stuff. That's simpler and should be used instead.

PR #5, opened 2014-05-17
Nodeispublished

rinasek:
Please review changes and merge if OK.

Thanks


src/Plugin/Condition/NodeIsPublished.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Provides a 'Node Is Published' condition.
+ *
+ * @Condition(
+ *   id = "rules_node_is_published",
+ *   label = @Translation("Node is published"),
+ *   context = {
+ *     "node" = {
+ *       "type" = "entity:node"
```
fago, line 21: This misses a label, see UserIsBlocked in the latest 8.x-3.x code for an example.

PR #6, opened 2014-05-17
Implementation of NodeIsPromoted Condition.

omissis:
Implementation for Ticket https://drupal.org/node/2269565


src/Plugin/Condition/NodeIsPromoted.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Drupal\rules\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Condition\ConditionInterface;
+
+use Drupal\Core\Executable\ExecutableManagerInterface;
+
+/**
+ * Defines a rule condition form promoted content.
```
fago: form? That description needs to be fixed.

src/Plugin/Condition/NodeIsPromoted.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @Condition(
+ *   id = "rules_node_is_promoted",
+ *   label = @Translation("Node is promoted"),
+ *   context = {
+ *     "node" = {
+ *       "type" = "entity:node",
```
fago: I wonder that's parsing? (Maybe d8 updated the annotation reader meanwhile).

src/Plugin/Condition/NodeIsPromoted.php
```diff
@@ -0,0 +1,44 @@
...
+ *   }
+ * )
+ */
+class NodeIsPromoted extends ConditionPluginBase implements ConditionInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return t('Content is promoted.');
```
fago: Do not add the trailing point, as it's not in the label either.

omissis:
@fago here's an updated version sporting both (phpunit) unit tests and (simpletest) integration tests, plus the fixes to strings and annotations.

src/Plugin/Condition/NodeIsPromoted.php
```diff
@@ -0,0 +1,44 @@
...
+ *   }
+ * )
+ */
+class NodeIsPromoted extends ConditionPluginBase implements ConditionInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return t('Content is promoted');
```
fago, line 34: This should match the label -> Node is promoted

src/Tests/Plugin/Condition/NodeIsPromotedIntegrationTest.php
```diff
@@ -0,0 +1,56 @@
...
+    );
+  }
+
+  public function setUp() {
+    parent::setUp();
+
+    $this->installSchema('node', array('node', 'node_field_data', 'node_field_revision', 'node_revision'));
+
+    // Create the node bundles required for testing.
+    $type = entity_create('node_type', array('type' => 'page', 'name' => 'page'));
```
fago, line 37: Instead of entity_create(), we should use the entity manager which is setuped by the EntityUnitTestBase already.
-> getStorage('node_type')->create($values)

src/Tests/Plugin/Condition/NodeIsPromotedIntegrationTest.php
```diff
@@ -0,0 +1,56 @@
...
+    $type->save();
```
fago, line 38: Does that work without save also? If so, let's do - else it would be good to comment why we need it.

src/Tests/Plugin/Condition/NodeIsPromotedIntegrationTest.php
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * Tests creating a rule and iterating over the rule elements.
```
fago, line 42: outdated

src/Tests/Plugin/Condition/NodeIsPromotedIntegrationTest.php
```diff
@@ -0,0 +1,56 @@
...
+   */
+  public function testConditionCreation() {
+    $manager = $this->container->get('plugin.manager.condition', $this->container->get('container.namespaces'));
+    $page = entity_create('node', array('type' => 'page', 'title' => $this->randomName()));
+
+    $condition = $manager->createInstance(self::CONDITION_ID)
+      ->setConfig('bundles', array('page'))
```
fago, line 49: No such config.

tests/src/Plugin/Condition/NodeIsPromotedTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+use Drupal\Tests\UnitTestCase;
+use Drupal\rules\Plugin\Condition\NodeIsPromoted;
+use Drupal\node\Entity\Node;
+
+/**
+ * Tests the rules' condition about promoted content flag.
+ */
+class NodeIsPromotedTest extends UnitTestCase {
+  const CONDITION_ID = 'rules_node_is_promoted';
```
fago, line 20: Once we do already the integration test, I don't think it's valuable the keep the unit test also. We need a simple pattern for devs on how to create conditions, which should involve creating only one test class. Therefore, let's go with the integration class only and check for the correct functionality there also.

fubhy:
Just a side-note: I think we should squash commits in these pull requests to reduce the number of "fixing this" / "whoops, typo!" / etc. commits ending up in the rules git log. Github automatically updates the pull request if you squash your commits.

omissis:
hi: you're right, thanks for the heads up.

fubhy:
There is now a duplication of this pull request at https://github.com/fago/rules/pull/29. It's basically a copy of the code here minus the unit test.

klausi:
Merged in the other pull request.

PR #11, opened 2014-05-27
Adding sphinx integration for documentation.

fubhy:
readthedocs.org is an awesome free service for building and hosting documentation generated from repositories. It uses Sphinx and with the configuration in this pull request it's going to be really simple for us to provide awesome documentation in reStructuredText format.


fubhy:
I set up a readthedocs.org account to configure a demo for this pull request: http://rules.readthedocs.org/en/latest/#

It's rather empty of course (no documentation yet :P) but a quick look at other documentation repositories hosted on that site should give you a good idea of how it might look.

@bojanz just recently linked this on Twitter: http://django-oscar.readthedocs.org/en/latest/topics/key_questions.html if you are looking for a good example ;)

fubhy:
I am going to put this into a separate repository (rules-docs). We should put the detailed documentation (including contrib doc) there and reference it from here I think. That ensures that our main repo stays clear from documentation artifacts like those static files (e.g. images) that we got in here now.

PR #15, opened 2014-05-28
Enabling tests for conditions.

fubhy:
This is going to cause fails because the 'Node is published' test is corrupt.


fubhy:
Aha.. Okay, so it doesn't fail because NodeIsPublishedTest is currently named "NodeIsPubishedTest" :) (without the l)

PR #16, opened 2014-05-28
Fixing node condition tests.

fubhy:
Those can't work currently without a) enabling the node module and b) actually saving the node type that they create.

### DIFF
--- a/src/Tests/Condition/NodeIsStickyTest.php
+++ b/src/Tests/Condition/NodeIsStickyTest.php
@@ -34,13 +34,6 @@ class NodeIsStickyTest extends EntityUnitTestBase {
   protected $nodeStorage;
 
   /**
-   * The node type used for testing.
-   *
-   * @var \Drupal\node\Entity\NodeType
-   */
-  protected $nodeType;
-
-  /**
    * {@inheritdoc}
    */
   public static function getInfo() {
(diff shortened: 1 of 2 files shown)


fago:
Why is saving the node type required? Why do the tests pass already then?

fubhy:
The tests don't pass currently. We did not test rules conditions before because of lacking configuraiton for that in the .travis.yml. And once we enabled it, we did not test the node condition for IsPublished (which was the only one back then) because the class was named incorrectly. Hence, no tests were actually firing.

fubhy:
The node type needs to be saved because we get an exception if we try to create a node later on that tries to use that node type. I did not dig deeper.... But it fails :)

src/Tests/Condition/NodeIsStickyTest.php
```diff
@@ -56,11 +49,12 @@ public static function getInfo() {
...
   public function setUp() {
     parent::setup();
-    $entity_manager = $this->container->get('entity.manager');
     $this->conditionManager = $this->container->get('plugin.manager.condition', $this->container->get('container.namespaces'));
-    $this->nodeStorage = $entity_manager->getStorage('node');
-    $this->nodeType = $entity_manager->getStorage('node_type')
-      ->create('node_type', array('type' => 'page'));
+    $this->nodeStorage = $this->container->get('entity.manager')->getStorage('node');
+
+    $this->container->get('entity.manager')->getStorage('node_type')
```
fago: There should work via $this->entityManager I think.

src/Tests/Condition/NodeIsStickyTest.php
```diff
@@ -56,11 +49,12 @@ public static function getInfo() {
    */
   public function setUp() {
     parent::setup();
-    $entity_manager = $this->container->get('entity.manager');
     $this->conditionManager = $this->container->get('plugin.manager.condition', $this->container->get('container.namespaces'));
-    $this->nodeStorage = $entity_manager->getStorage('node');
-    $this->nodeType = $entity_manager->getStorage('node_type')
-      ->create('node_type', array('type' => 'page'));
+    $this->nodeStorage = $this->container->get('entity.manager')->getStorage('node');
```
fago: here as well

PR #17, opened 2014-05-28
Adding a simple 'user has role' condition.

fubhy:


fubhy:
This does not resemble the condition from D7 100% because it apparently operates on a list of roles and can be either evaluated with AND or OR. I wonder if we still want that?

fago:
Please see https://drupal.org/node/1921540 - I guess we better participate there.

fubhy:
Ah okay... Didn't know that exists.

PR #21, opened 2014-05-29
Action expression.

klausi:
See https://drupal.org/node/2276279


src/Plugin/RulesExpression/ActionExpression.php
```diff
@@ -0,0 +1,89 @@
...
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Action\ActionManager $actionManager
+   *   The action manager.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, \Drupal\Core\Action\ActionManager $actionManager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
```
fago, line 55: Let's make the $configuration array the same as the d7 settings array of rules elements. But in addition let's make a nice and fluid API to change that configruation instead of d7 public access.

klausi:
Changed that to use the configuration array and pushed to the 8.x-3.x branch.

PR #23, opened 2014-05-29
Enabling web test support.

fubhy:


.travis.yml
```diff
@@ -50,8 +51,8 @@ before_script:
 
   # start a web server on port 8080, run in the background; wait for initialization
   # This fails currently because there is a missing drush dependency.
```
klausi: that comment should be removed, right?

fubhy:
Umh.. Right :)

PR #27, opened 2014-05-29
Feature/context

fago:
Issue #2245049: Implement a Rules variant of the core context system to improved DX and have a context definition class that is extensible with further information as needed by Rules.

### DIFF
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -7,6 +7,8 @@
 
 namespace Drupal\rules\Context;
 
+use Drupal\Core\Entity\ContentEntityBase;
+use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\TypedData\TypedDataInterface;
 
 /**
@@ -39,6 +41,12 @@ class Context implements ContextInterface {
    * {@inheritdoc}
    */
   public function getContextValue() {
+    // Special case entities.
+    // @todo: Remove once entities do not implemented TypedDataInterface any
+    // more.
+    if ($this->contextData instanceof ContentEntityInterface) {
+      return $this->contextData;
+    }
     return $this->contextData->getValue();
   }
 
@@ -46,7 +54,12 @@ class Context implements ContextInterface {
    * {@inheritdoc}
    */
   public function setContextValue($value) {
-    return $this->setContextData(\Drupal::typedDataManager()->create($this->contextDefinition->getDataDefinition(), $value));
+    if ($value instanceof TypedDataInterface) {
+      return $this->setContextData($value);
+    }
+    else {
+      return $this->setContextData(\Drupal::typedDataManager()->create($this->contextDefinition->getDataDefinition(), $value));
+    }
   }
 
   /**
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -7,7 +7,6 @@
 
 namespace Drupal\rules\Context;
 
-use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\TypedData\TypedDataInterface;
 
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -1,0 +1,100 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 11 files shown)


src/Context/Context.php
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setContextValue($value) {
+    if ($value instanceof TypedDataInterface) {
+      return $this->setContextData($value);
+    }
+    else {
```
omissis, line 59: no need for else here.

src/Context/ContextAwarePluginInterface.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+namespace Drupal\rules\Context;
+
+use \Drupal\Component\Plugin\ContextAwarePluginInterface as CoreContextAwarePluginInterface;
+
+/**
+ * Interface for Rules-context aware plugins.
+ */
+interface ContextAwarePluginInterface extends CoreContextAwarePluginInterface {
```
omissis, line 15: I am not sure I understand why most of the functions that already declared in CoreContextAwarePluginInterface are redeclared here, but I guess it's because you need to return different types than the parent interface. I don't think that's a safe idea: that would make the two interfaces not swappable and basically makes them hidden-ly incompatible.

src/Context/ContextDefinition.php
```diff
@@ -0,0 +1,190 @@
...
+   */
+  public function addConstraint($constraint_name, $options = NULL) {
+    $this->constraints[$constraint_name] = $options;
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDataDefinition() {
```
omissis, line 175: I would probably consider a ContextDefinitionCollection class to deal with multiple context data definitions and avoid conditional logic here. Also I'd inject the typed data manager instead of calling it from inside the method.

src/Engine/RulesConditionBase.php
```diff
@@ -0,0 +1,117 @@
...
+   */
+  public function submitConfigurationForm(array &$form, array &$form_state) {
+    // @todo: Figure out whether this is useful to Rules somehow.
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    return $this->executableManager->execute($this);
```
omissis, line 83: What happens if executableManager is not set? maybe it's safer to move the injection from setter to constructor if this method's executed often.

src/Engine/RulesConditionBase.php
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getConfiguration() {
+    return array(
+      'id' => $this->getPluginId(),
+    ) + $this->configuration;
```
omissis, line 92: Is $this->configuration always an guaranteed to be an array?

src/Context/ContextAwarePluginInterface.php
fago, line 15: Yes, the goal would be to get our interface in core. For now, that's what we can do easily.

src/Context/ContextDefinition.php
fago, line 175: Problem with injecting is that this is a definition object that's intended to easily serializable. That's an unsolved problem right now in core also.

src/Engine/RulesConditionBase.php
fago, line 83: Good point. That's probably something we should do, not necessarily part of this PR though. This is actually just copied over from the core-code, so we do not have to use the core condition base class. Would you like to create an issue for it? :)

src/Engine/RulesConditionBase.php
fago, line 92: yep, see setConfiguration.

omissis, line 92: yeah that's exactly why I asked: the presence of a setter does not guarantee that a property is set. But I noticed now that the grandparent's constructor requires and set configuration. That's not 100% safe because a child class can still override the constructor thus breaking configuration, but I think it's ok.

PR #7, opened 2014-05-18
#2269575: Implement "Path has url alias" condition

Nebel54:

PR #31, opened 2014-06-03
#2269575 - PathHasAlias condition implementation. 

paranojik:
Another take at implementing the PathHasAlias condition. Previously opened as https://github.com/fago/rules/pull/7.

### DIFF
--- a/src/Plugin/Condition/PathHasAlias.php
+++ b/src/Plugin/Condition/PathHasAlias.php
@@ -1,0 +1,97 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\PathHasAlias.
+ */
+
+namespace Drupal\rules\Plugin\Condition;
+
+use Drupal\Core\Path\AliasManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Engine\RulesConditionBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Path has alias' condition.
+ *
+ * @Condition(
+ *   id = "rules_path_has_alias",
+ *   label = @Translation("Path has alias")
+ * )
+ *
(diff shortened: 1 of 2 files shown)


src/Plugin/Condition/PathHasAlias.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\PathHasAlias.
+ */
+
+namespace Drupal\rules\Plugin\Condition;
+
+
```
fubhy: Please remove the extra line break.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -0,0 +1,55 @@
...
+use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Engine\RulesConditionBase;
+
```
fubhy: Same here.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -0,0 +1,55 @@
...
+    return $this->t('Path has alias');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $path = $this->getContextValue('path');
+    $language = $this->getContextValue('language');
+    return \Drupal::service('path.alias_storage')->lookupPathAlias($path, $language) !== FALSE;
```
fubhy: Please inject the path alias storage service properly using the static create method.

paranojik:
This is now almost completely copied from PathAliasExists condition. Still don't know why tests are failing...

fubhy:
I can a look if you don't mind.

paranojik:
Sure. Go ahead.

2014-06-05 10:08 GMT+02:00 fubhy notifications@github.com:

fubhy:
Continued in https://github.com/fago/rules/pull/35

The reason for the failures here: There is something called AliasWhitelist in Drupal core that ensures that you cannot create/lookup path aliases for paths that don't actually exist. Hence you need to use a path like 'user' (which exists once you enable the system module which inheretly enables the user module).

PR #4, opened 2014-05-17
Added Path Alias Exists condition

bbujisic:
There was an error with doctrine. The new pull request fixed it.

PR #32, opened 2014-06-04
Added 'Path alias exists' condition.

fubhy:
Continued from https://github.com/fago/rules/pull/4

Credit going to Branislav (@calcanevs) of course.

### DIFF
--- a/src/Plugin/Condition/PathAliasExists.php
+++ b/src/Plugin/Condition/PathAliasExists.php
@@ -1,0 +1,98 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\PathAliasExists.
+ */
+
+namespace Drupal\rules\Plugin\Condition;
+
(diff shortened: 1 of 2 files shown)


src/Plugin/Condition/PathAliasExists.php
```diff
@@ -0,0 +1,99 @@
...
+    return $this->t('URL alias exists');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $alias = $this->getContextValue('alias');
+    $language = $this->getContext('language')->getContextData() ? $this->getContextValue('language') : NULL;
+    $language = $language ? $language->getId() : NULL;
```
klausi: don't resuse the variable, call it lang_code for clarity.
I think you don't even need the two lines? just call getId() on the first line if the context exists?

fubhy: Yeah, just didn't want to make the line longer than necessary ;)

PR #33, opened 2014-06-04
Added mocked test rule to unit test base.

fubhy:
That's better than tricking the constructor by giving it bogus arguments.


fubhy:
Was that a miss click or is something wrong with this? :)

PR #34, opened 2014-06-04
Use mocked 'and' condition container

fubhy:
This is better than manually creating one, bypassing the constructor with bogus arguments.


fubhy:
Okay... Please review now :)

PR #35, opened 2014-06-05
Added a 'Path has alias' condition.

fubhy:


paranojik:
So @fubhy what did you change to make it work?
btw: you left a reference to "en" in testConditionEvaluationLanguageNeutral()  Line 79 of src/Tests/Condition/PathHasAliasTest.php.

fubhy:
Yep, because that tests the fallback (e.g. when requesting an alias for 'en' and it does not exist but a language neutral one does, it should still return the language neutral one).

I added a comment on the old PR that explains what I changed:

https://github.com/fago/rules/pull/31

"The reason for the failures here: There is something called AliasWhitelist in Drupal core that ensures that you cannot create/lookup path aliases for paths that don't actually exist. Hence you need to use a path like 'user' (which exists once you enable the system module which inheretly enables the user module)."

paranojik:
Stupid me... I noticed that afterwards...sorry. Thanks for the explanation.
I struggled to find documentation on that topic, but all I was able to find
was an outdated change notice.
Anyway, thanks again.

2014-06-05 23:08 GMT+02:00 fubhy notifications@github.com:

fubhy:
Yeah, no worries. Thanks for trying :). I spent about 15 minutes in XDebug too before I found the problem.

PR #36, opened 2014-06-05
Added a trait for expression creation methods.

fubhy:


src/Tests/RulesEngineTest.php
```diff
@@ -29,18 +29,31 @@ public static function getInfo() {
...
-        ->addCondition($this->createExpression('rules_and')
-          ->addCondition($this->createCondition('rules_test_false'))
-          ->addCondition($this->createCondition('rules_test_true')->negate())
-          ->negate()));
-    $rule->addAction($this->createAction('rules_test_log'));
+    // Create an 'and' condition container and add conditions to it.
+    $and = $this->createRulesAnd()
+      ->addCondition($this->createRulesCondition('rules_test_false'))
+      ->addCondition($this->createRulesCondition('rules_test_true')->negate())
+      ->negate();
```
klausi, line 36: Could you also include the result of the and in the comment like "The 'and' evaluates to TRUE"?

src/Tests/RulesEngineTest.php
```diff
@@ -29,18 +29,31 @@ public static function getInfo() {
...
+
+    // Create an 'or' condition container and add conditions to it, including
+    // the previously created 'and' condition container.
+    $or = $this->createRulesOr()
+      ->addCondition($this->createRulesCondition('rules_test_true')->negate())
+      ->addCondition($this->createRulesCondition('rules_test_false'))
+      ->addCondition($and);
```
klausi, line 46: Same here, it would be nice to know the intermediate outcome of this OR in the comment.

fubhy:
Regarding your comments... What about we execute these intermediate steps too to show that? Like... add assertions for those?

klausi:
Sure, that would make it easier to debug this test and show at what step it went wrong. Good idea!

fubhy:
Done.

PR #38, opened 2014-06-05
Added unit tests for RulesExpressionTrait.

fubhy:


tests/src/RulesExpressionTraitTest.php
```diff
@@ -0,0 +1,126 @@
...
+      ->setMethods(['createInstance'])
+      ->disableOriginalConstructor()
+      ->getMock();
+
+    $this->expressionManager->expects($this->any())
+      ->method('createInstance')
+      ->will($this->returnCallback(function ($id) {
+        $type = ucfirst(substr($id, 6));
+        return $this->{"getMock$type"}();
+      }));
```
klausi, line 66: WTH is going on here? Please add a comment and an example what gets converted into what.

tests/src/RulesExpressionTraitTest.php
```diff
@@ -0,0 +1,126 @@
...
+   * @covers ::createRulesAnd()
+   * @covers ::createRulesOr()
+   *
+   * @dataProvider convenienceMethodsDataProvider
+   *
+   * @todo Test ::createRulesAction() and ::createRulesCondition().
+   */
+  public function testConvenienceMethods($id, $method, $class) {
+    // Test the 'createRulesExpression' method.
+    $reflection = $this->reflection->getMethod('createRulesExpression');
```
klausi, line 103: $reflection is a bad name, would $method be better?

tests/src/RulesExpressionTraitTest.php
fubhy, line 66: Hehe, Okay... Will add that tomorrow :)

tests/src/RulesExpressionTraitTest.php
fubhy, line 103: $method is already in use. We can name it $reflection_method or name the other variable $method_name.

fubhy:
I simplified the unit test in this commit: https://github.com/fago/rules/commit/1da91da505722d28a00731cb7c44d8be1b04b8c8

PR #40, opened 2014-06-06
Use injected typed data manager in Context API.

fubhy:


src/Context/ContextAwarePluginBase.php
```diff
@@ -31,14 +33,77 @@
   protected $contextDefinitions;
 
   /**
+   * The typed data manager.
```
klausi: What is the typed data manager used for in this class?

src/Context/ContextDefinition.php
```diff
@@ -43,22 +45,41 @@ class ContextDefinition implements ContextDefinitionInterface {
   protected $constraints = [];
 
   /**
+   * The typed data manager.
```
klausi: Same here, why do we need the manager?

PR #46, opened 2014-06-06
Added system message action and rules action interface.

klausi:


src/Plugin/Action/DrupalMessage.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\Core\TypedData\TypedDataManager;
+use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Engine\RulesActionBase;
+
+/**
+ * Provides an action to show a message on the site.
```
fubhy: let's write this like we did with the conditions 'Provides a "Show a message on the site" action.'

PR #51, opened 2014-06-06
Issue #2281111: Porting 'Entity is of bundle (entity_is_of_bundle)' cond...

ndewhurst:
...ition to 8.x

See: d.o issue #2281111 - https://drupal.org/node/2281111
This creates an EntityIsOfBundle condition and an associated test that checks a simulated node entity's type and bundle.

### DIFF
--- a/src/Plugin/Condition/EntityIsOfBundle.php
+++ b/src/Plugin/Condition/EntityIsOfBundle.php
@@ -7,7 +7,6 @@
 
 namespace Drupal\rules\Plugin\Condition;
 
-use Drupal\Core\TypedData\TypedData;
 use Drupal\Core\TypedData\TypedDataManager;
 use Drupal\rules\Context\ContextDefinition;
 use Drupal\rules\Engine\RulesConditionBase;
@@ -25,22 +24,6 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  * @todo: Add group information from Drupal 7?
  */
 class EntityIsOfBundle extends RulesConditionBase {
-
-  /**
-   * Constructs an EntityIsOfBundle object.
-   *
-   * @param array $configuration
-   *   A configuration array containing information about the plugin instance.
-   * @param string $plugin_id
-   *   The plugin ID for the plugin instance.
-   * @param mixed $plugin_definition
-   *   The plugin implementation definition.
-   * @param \Drupal\Core\TypedData\TypedDataManager $typed_data_manager
-   *   The typed data manager.
-   */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, TypedDataManager $typed_data_manager) {
-    parent::__construct($configuration, $plugin_id, $plugin_definition, $typed_data_manager);
-  }
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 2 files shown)


src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\TypedData\TypedDataManager $typed_data_manager
+   *   The typed data manager.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, TypedDataManager $typed_data_manager) {
```
fago: Does nothing different than the parent, thus needs to go away.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function contextDefinitions(TypedDataManager $typed_data_manager) {
+    $contexts['entity'] = ContextDefinition::create($typed_data_manager, 'entity')
+      ->setLabel(t('Entity'))
+      ->setDescription(t('Specifies the entity for which to evaluate the condition.'));
+
+    // @todo: specify input mechanisms and/or constraints once configuration/UI
```
fago: Constraints we can specify already, so it should be done now. Also, let's just add todo line for each todo item + start each sentence uppercase end end it with a trailing point.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    //  questions are settled.
+    $contexts['type'] = ContextDefinition::create($typed_data_manager, 'string')
+      ->setLabel(t('Type'))
+      ->setDescription(t('The type of the evaluated entity.'));
+
+    // @todo: specify input mechanisms and/or constraints once configuration/UI
```
fago: same as above

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    //  questions are settled.
+    $contexts['bundle'] = ContextDefinition::create($typed_data_manager, 'string')
+      ->setLabel(t('Bundle'))
+      ->setDescription(t('The bundle of the evaluated entity.'));
+
+    return $contexts;
+  }
+
```
fago, line 62: one empty line too many

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+   */
+  public function summary() {
+    return $this->t('Entity is of bundle');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    // Load the entity to evaluate.
```
fago: unnecessary comment, reading the code tells me that.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    $providedEntity = $this->getContextValue('entity');
+    // Retrieve the entity type specified as part of this condition.
```
fago: unnecessary comment, reading the code tells me that.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    $specified_type = $this->getContextValue('type');
+    // Retrieve the entity bundle specified as part of this condition.
```
fago: unnecessary comment, reading the code tells me that.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    $specified_bundle = $this->getContextValue('bundle');
+    // Retrieve the type and bundle of the evaluated entity.
```
fago: unnecessary comment, reading the code tells me that.

src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -0,0 +1,105 @@
...
+    $entity_type = $providedEntity->getEntityTypeId();
+    $entity_bundle = $providedEntity->bundle();
+    // Check to see whether the entity's bundle and type match the specified
+    //  values.
```
fago, line 80: one white space too many

src/Plugin/Condition/EntityIsOfBundle.php
ndewhurst: OK, cool. I'm not familiar with constraints, but the concept makes sense. It looks like this is the authoritative reference for available constraints: https://api.drupal.org/api/drupal/namespace/Symfony%21Component%21Validator%21Constraints/8
What constraints would be relevant here? "NotBlank"...or "NotNull" may be relevant for all three. "Type" seems redundant. "Length" would make sense for the strings...

fubhy:
Thanks, merged! I swapped the simple test out for a unit test while merging.

PR #55, opened 2014-06-06
Proof of concept for condition unit tests with proper annotation discovery

fubhy:


tests/src/Condition/NodeIsPublishedTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Tests the context definitions.
+   *
+   * @covers ::contextDefinitions()
+   */
+  public function testContextDefinition() {
+    // Test that the 'path' context is properly defined.
+    $context = $this->condition->getContext('node');
+    $this->assertInstanceOf('Drupal\rules\Context\ContextInterface', $context);
```
fago, line 75: Those checks seem to be rather verbose. Maybe we can provide a helper that does general context assertions given a context key.

fubhy, line 75: Sure. This is not the important part of this PR :P. This is a proof of concept for the discovery!

tests/src/Mock/ContainerFactory.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace Drupal\rules\Tests\Mock;
+
+use Drupal\Core\Plugin\Factory\ContainerFactory as CoreContainerFactory;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Mock container factory for testing plugins.
```
fago, line 15: Why do we need this. We should clearly document for each mock class why it's needed.

fubhy:
I don't think that this is a good idea anymore. It's mixing integration and unit test and there is a high probability for it breaking if anything else in a given Drupal setup modifies anything with the annotation or adds custom stuff.

PR #57, opened 2014-06-08
Entity field access condition 2281113

torenware:
This is a first pass on #2281113.  Plugin is implemented and the test runs green, but only hits the "allowed" case.  Other questions:
- How should language be handled?  I'm taking the "default language", and I'm not sure that's right.
- I don't test the "not allowed" case.  I'd guess that I should modify the test module, and make SimpleTest include it.  Is this what I should do?
- I need to check coding standards, since it looks like we do this differently in 8.x.
- I worked to get this working;  it probably needs to be changed to correctly support DI.

### DIFF
--- a/src/Engine/RulesActionContainerInterface.php
+++ b/src/Engine/RulesActionContainerInterface.php
@@ -7,22 +7,20 @@
 
 namespace Drupal\rules\Engine;
 
-use Drupal\Core\Action\ActionInterface;
-
 /**
  * Defines a common interface for action containers.
  */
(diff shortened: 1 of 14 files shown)


torenware:
Hold a bit on reviewing this;  I just noticed I'm testing ::access() and not ::fieldAccess().  This should be fixed in the next hour or so.

torenware:
This issue is supplanted by https://github.com/fago/rules/pull/61 (I had problems rebasing the history)

PR #60, opened 2014-06-08
Clean up the documentation blocks.

bartfeenstra:
FIX ALL THE THINGS.

### DIFF
--- a/src/Context/ContextAwarePluginBase.php
+++ b/src/Context/ContextAwarePluginBase.php
@@ -85,7 +85,6 @@ abstract class ContextAwarePluginBase extends PluginBase implements ContextAware
    *   The typed data manager.
    *
    * @return $this
-   *   The current object for chaining.
    */
   public function setTypedDataManager(TypedDataManager $typedDataManager) {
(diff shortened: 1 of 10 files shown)


fubhy:
Are those changes according to CS? My Drupal CS knowledge is a bit rusty (didn't do much core lately).

bartfeenstra:
Yes. We don't need descriptions for @return $this, as this is 
self-explanatory (it _always_ returns the object the method is called 
on). I fixed some spelling errors, and rephrased some comments a bit. 
Interfaces don't actually provide things, but define them.

fubhy:
If "this is self-explanatory" would be enforced all across core our CD for comments would like rather differently :P. (I hate how we keep documenting absolutely redundant stuff).

bartfeenstra:
I don't know what you mean, but @return $this without a description is 
part of the coding standards.

fubhy:
Yeah, I meant other things. We tend to over-document self-explanatory @vars and constructor arguments and things like that... Not part of this here though.

PR #61, opened 2014-06-09
Entity field access condition 2281113b

torenware:
Updated version of my previous patch;  I've changed branches due to some very bad problems with "git rebase".

I'm aware I have two commits that appear here.  That's the problem -- I can't get git to go lower without pulling in commits I don't own.  If you folks can tell me how to fix this, much thanks.

The patch creates the plugin and test for the issue.

### DIFF
--- a/src/Plugin/Condition/UserHasEntityFieldAccess.php
+++ b/src/Plugin/Condition/UserHasEntityFieldAccess.php
@@ -1,0 +1,122 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\UserHasEntityFieldAccess.
+ */
+
+namespace Drupal\rules\Plugin\Condition;
+
+use Drupal\Core\TypedData\TypedDataManager;
+use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Engine\RulesConditionBase;
+use Drupal\Core\Language\Language;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'User has entity field access' condition.
+ *
+ * @Condition(
+ *   id = "rules_entity_field_access",
+ *   label = @Translation("User has entity field access")
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add group information from Drupal 7.
+ */
+class UserHasEntityFieldAccess extends RulesConditionBase {
+
+  /**
+   * The entity manager.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+  
+  /**
+   * Constructs a UserHasEntityFieldAccess object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\TypedData\TypedDataManager $typed_data_manager
+   *   The typed data manager.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, TypedDataManager $typed_data_manager, EntityManagerInterface $entity_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition, $typed_data_manager);
+    $this->entityManager = $entity_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('typed_data_manager'),
+      $container->get('entity.manager')
+    );
+  }
+  
+  /**
+   * {@inheritdoc}
+   */
+  public static function contextDefinitions(TypedDataManager $typed_data_manager) {
+    $contexts['entity'] = ContextDefinition::create($typed_data_manager, 'entity')
+      ->setLabel(t('Entity'));
+
+    $contexts['field'] = ContextDefinition::create($typed_data_manager, 'string')
+      ->setLabel(t('Field Name'));
+
+    $contexts['op'] = ContextDefinition::create($typed_data_manager, 'string')
+      ->setLabel(t('Operation'));
+
+    $contexts['account'] = ContextDefinition::create($typed_data_manager, 'entity:user')
+      ->setLabel(t('User'));
+
+    return $contexts;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t('User has access to field on entity');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $entity = $this->getContextValue('entity');
+
+    $field = $this->getContextValue('field');
+    if (!$entity->hasField($field)) {
+      return FALSE;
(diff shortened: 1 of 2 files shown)


src/Plugin/Condition/UserHasEntityFieldAccess.php
```diff
@@ -0,0 +1,102 @@
...
+  public function evaluate() {
+    $entity = $this->getContextValue('entity');
+    $field = $this->getContextValue('field');
+    if (!$entity->hasField($field)) {
+      return FALSE;
+    }
+    $field_definition = $entity->getFieldDefinition($field);
+    $op = $this->getContextValue('op');
+    $account = $this->getContextValue('account');
+    $entity_access_controller = \Drupal::entityManager()->getAccessController($entity->getEntityTypeId());
```
fubhy: Please use the static ::create() method to inject the access controller.

src/Plugin/Condition/UserHasEntityFieldAccess.php
```diff
@@ -0,0 +1,102 @@
...
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t('User has access to field on entity');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
```
fubhy, line 99: Please add some line-breaks in this methods to separate the logical blocks. It's kinda hard to read :)

src/Tests/Condition/UserHasEntityFieldAccessTest.php
```diff
@@ -0,0 +1,169 @@
...
+      'status' => 1,
+    ]);
+  }  
+
+  /**
+   * Tests evaluating the condition.
+   */
+  public function testConditionEvaluation() {
+    $condition = NULL;
+    try {
```
fubhy: Why is that necessary?

src/Plugin/Condition/UserHasEntityFieldAccess.php
torenware, line 99: I had read the new style guidelines, which seemed to say I should remove them :-)

But I agree, white space is good.  I'll add it.

src/Tests/Condition/UserHasEntityFieldAccessTest.php
torenware: Are you referring to the "$condition = NULL" section?

I am trying to test for plugin discovery.  But it does not currently work as I expected;  having $this->assertNotNull($condition) does not end execution in my tests when $condition is NULL, and execution continues into "$condition->setContext('node..." where it dies.

I can certainly remove that if you don't think it's needed.

If it's for the getUser() section:  when looking for a utility to create a user directly, I did not find EntityUnitTestBase::createUser(), which does appear to do what I need -- just found it now.  But I can easily use that instead.

Related to that:  can I extend the rules_test module to create a field with access control, via a permission?  This would be a lot easier than try to create the access control in the test itself, since all I'll need to do is attach the field, and assign the user to a permission defined in rules_test?

I can make that change as part of this issue, or as a separate issue, however you see appropriate.

src/Plugin/Condition/UserHasEntityFieldAccess.php
torenware: Easily done.  Will add that to ::create and to the constructor.

torenware: fubhy --

It turns out that EntityAccessControllerInterface is not currently exposed via a service, so the $container won't know about it.  Currently, there _is_ a service for entity.manager, and I'll code to include it.

The access control can't really be injected directly right now, since each entity type has its own controller.

src/Tests/Condition/UserHasEntityFieldAccessTest.php
torenware: It appears that the entity_test module defined by the System module has enough bells and whistles.  I'll change my test to use that entity type, instead of the entity:node:page entity I'm using now.

src/Plugin/Condition/UserHasEntityFieldAccess.php
fubhy: Correct, the access handler itself is not a service, but it is registered with one, namely the EntityManager (which is a service). So what you do is (in ::create). $container->get('entity.manager')->getAccessController('your-entity-type').

src/Tests/Condition/UserHasEntityFieldAccessTest.php
fubhy: Don't worry so much about the test! All I want for that at this point is for it to show me that the plugin itself works. It doesn't have to be perfect because we are going to replace it with a unit test really soon. If you feel comfortable with unit tests, you can take a look at our other Condition plugin unit tests in rules/tests/src/Conditions (take the latest checkout from 8.x-3.x to see a set of which I already converted).

Really not required though. We can work through this together if you like (using simpletest) or you can give Unit tests a shot (which I am going to help you with too then). The existing unit tests should give you a good idea of what's going on.

fubhy:
Your current checkout is still dirty... Let me create a fresh branch for that on my fork. I will send you a link to that in a second so that you can delete your local branch, take my clean one, and then force push it to your fork.

fubhy:
... Or probably even better: I can show you how you can fix your branch.

fubhy:
I created a clean branch with your changes in https://github.com/fubhy/rules/tree/entity-field-access-condition-2281113b

Please put that back into this PR. You can do that like so:

1.) Add my fork as a remote (git remote add fubhy https://github.com/fubhy/rules.git)
2.) Fetch my repo (git fetch fubhy)
3.) Go to your branch (git checkout entity-field-access-condition-2281113b)
4.) Reset your branch to what's in mine (git reset --hard fubhy/entity-field-access-condition-2281113b)
5.) Force-push to your remote (git push origin entity-field-access-condition-2281113b --force)

src/Tests/Condition/UserHasEntityFieldAccessTest.php
torenware: > Don't worry so much about the test! All I want for that at this point is for it to show me that the plugin itself works. It doesn't have to be perfect because we are going to replace it with a unit test really soon. If you feel comfortable with unit tests, you can take a look at our other Condition plugin unit tests in rules/tests/src/Conditions (take the latest checkout from 8.x-3.x to see a set of which I already converted).
> 
> Really not required though. We can work through this together if you like (using simpletest) or you can give Unit tests a shot (which I am going to help you with too then). The existing unit tests should give you a good idea of what's going on.
> 
> —
> Reply to this email directly or view it on GitHub.

My mail server was down since Thursday, so I just saw this.

Too late! :-)  I kept at it, and I have a non-trivial test working.  Runs nice and green, actually tests both of the main cases.

I'll see about getting my tree fixed.  If I can't, I'll just create a new branch and copy the two files over.  This will guarantee that the tree is clean.

Rob

torenware:
fubhy --

I just saw your reordering of my "b" branch;  much thanks.

I'm still having problems with rewriting history.  Let's use #63 for now;  I've spent almost as much time trying to get my tree unscrewed as I have on the patch, and the patch is the important thing.  I'll see if I can read up more on "git rebase" so I can understand where I'm having trouble.

torenware:
Fubhy -- I've figured out how to squash the commit.  #61 should be also good to go.  Looks like you get credit for my work in this version.

Hopefully rebasing future commits won't be this hard.

src/Tests/Condition/UserHasEntityFieldAccessTest.php
torenware: > Really not required though. We can work through this together if you like (using simpletest) or you can give Unit tests a shot (which I am going to help you with too then). The existing unit tests should give you a good idea of what's going on.
> 
> —
> Reply to this email directly or view it on GitHub.

I actually know PHPUnit a lot better than SimpleTest (I'm using a back-ported version of the D8 system in a couple of D7 projects), so I'm fine with that. 

Just show me where the PHPUnit tests go, and what test class you want me to derive from.

Thanks,
Rob

PR #63, opened 2014-06-10
Issue #2281113: User has entity field access condition for 8.x-3.x

torenware:
Previous attempts got into a very bad git state, mixing in several sets of patches.  This branch is built from scratch, and only has one commit at the top.

This issue supercedes https://github.com/fago/rules/pull/61, and includes the changes fubhy recommended in that issue.  Hopefully this one will be good to go.

### DIFF
--- a/src/Plugin/Condition/UserHasEntityFieldAccess.php
+++ b/src/Plugin/Condition/UserHasEntityFieldAccess.php
@@ -1,0 +1,122 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\UserHasEntityFieldAccess.
+ */
+
+namespace Drupal\rules\Plugin\Condition;
+
+use Drupal\Core\TypedData\TypedDataManager;
+use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Engine\RulesConditionBase;
+use Drupal\Core\Language\Language;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'User has entity field access' condition.
+ *
+ * @Condition(
+ *   id = "rules_entity_field_access",
+ *   label = @Translation("User has entity field access")
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add group information from Drupal 7.
+ */
+class UserHasEntityFieldAccess extends RulesConditionBase {
+
+  /**
+   * The entity manager.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+  
+  /**
+   * Constructs a UserHasEntityFieldAccess object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\TypedData\TypedDataManager $typed_data_manager
+   *   The typed data manager.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, TypedDataManager $typed_data_manager, EntityManagerInterface $entity_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition, $typed_data_manager);
+    $this->entityManager = $entity_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('typed_data_manager'),
+      $container->get('entity.manager')
+    );
+  }
+  
+  /**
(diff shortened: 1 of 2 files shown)


fubhy:
Thanks! Merged. 

I fixed some minor issue on commit and swapped out the simpletest with a unit test.

torenware:
fubhy,

Thanks for getting the code in.

Some questions about the PHP Unit test.

I looked at the code, and ran them via phpunit as directed in the module README.  I've confirmed the new test is called (I added $this->assertTrue(FALSE, 'we were called')  into one of the tests,  and it failed with my error message), and that it does indeed invoke my plugin (which I determined in the debugger).

But once the mocks are put in, what exactly are we testing?  In the simpletest version of the test, there were real fields on a real entity, and I could tell that the Field API code was called, and that the code failed or succeeded where it should have.

AFAICT, the mocks never actually invoke any real code; it's all simulated.

So how do we know the test is "valid" -- i.e., once we invoke it in an environment where the code is integrated into a real application, do the tests really tell us our code is valid?

I get the value of using mocks., but I'm wondering just what we're testing here.  Part of what we need to test, certainly, is whether the code will actually integrate when we need it to.

Thanks,

Rob Thorne
Torenware Networks

fubhy:
We are testing the Plugin implementation. Most notably the "evaluate" method. A unit test is sufficient for these plugin classes. We have to expect and trust that the code that drives the other bits and pieces that we are invoking from that method to work properly. There are integration tests that test the rules engine, there are tests in core that test fields, entities, etc. and all we need to test here is whether our implementation of the evaluate() method invokes these other systems according to our expectations and returns the right values with the given input (context). That is absolutely sufficient and even preferable.

torenware:
On Jun 20, 2014, at 7:30 AM, fubhy notifications@github.com wrote:

> We are testing the Plugin implementation. Most notably the "evaluate" method.
> 
> That's what it looked like.
> A unit test is sufficient for these plugin classes. We have to expect and trust that the code that drives the other bits and pieces that we are invoking from that method to work properly.
> 
> Having spent a lot of time around testers, I'm not sure we _should_ trust other parts of the system.

The problem here is that Core can change in such a way that it breaks the test used in ::execute().  As we are currently coding the unit tests, we are testing the conformance of ::execute() to the definitions of their respective mocks.  Note however, that we are not verifying that the mocks behave like the real system.  If the mocks are incorrectly specified,  phpunit will run green, but the plugin will not actually work out in the wild.  

Since Core itself is a moving target, I do think we need to do some kind of test where we detect if Core has broken a plugin.  Since fago is currently negotiating on changes of the Condition interface and related details of Core, I'd expect that we'll need to deal with this in some cases:  we'll get changes in both the interfaces and the behaviors of some APIs.    

We also need to worry in some cases if we are doing the right test in our plug-ins.  In my case, the correct check right now has two separate steps.  I know this from reading the test for field access in Core.  But I could have easily misread this (in fact, I did get the test wrong in my first version of the plugin).  The mocks might well accept my implementation as valid when it would never work in the wild.  If the Condition is not so that it will integrated, we want that condition to fail on test.

Now, the unit test does not need to be the test code that checks for relevant changes in Core.  We could keep some SimpleTest code around to cover this.  Or we could have a phpunit base class that let us bootstrap Drupal that we would use for these purposes.  But we really should do one or the other.

> There are integration tests that test the rules engine, there are tests in core that test fields, entities, etc. and all we need to test here is whether our implementation of the evaluate() method invokes these other systems according to our expectations
> 
> What keeps us in sync with Core here if Core changes?  The mocks will read the interfaces (I believe this use introspection for this), but an interface only promises you that the calling conventions conform;  it won't guarantee that the individual methods behave as you might expect them to behave.
> and returns the right values with the given input (context). That is absolutely sufficient and even preferable.
> 
> I'd argue not, for the reasons above.

It's an important axiom in software QA that if you test the wrong thing, you haven't tested the feature.  It still appears that the mocks will let us build plug-ins, and "pass" them, when the plugins will not actual work when integrated.  This needs to worry us.

Cheers,
Rob

fubhy:
> Having spent a lot of time around testers, I'm not sure we _should_ trust other parts of the system

If we follow that paradigm (and drive it a little bit ad absurdum), we need to test all of core in Rules.

> Note however, that we are not verifying that the mocks behave like the real system.  If the mocks are incorrectly specified,  phpunit will run green, but the plugin will not actually work out in the wild.
> If the tests are incorrect then they might pass even though the code is wrong. 

Correct. That is also the case with anything other than unit tests though.

> The problem here is that Core can change in such a way that it breaks the test used in ::execute().

If the tests break due to a change in core, it means that the implementation is very likely to break too. Our mocks of the core system represent the interfaces/classes used in our plugin implementations. All we need to know is if our plugin implementation uses the other systems properly according to their interfaces. If the interface changes or the plugin is doing something wrong, it should fail. Otherwise it should pass. That's all we need to test. Anything else would suggest that unit tests are a myth and would never apply because "we have to test the whole thing all the time" which is obviously not possible. We have to trust that Core and other providing systems have their own tests to cover the interfaces and classes that we use. We just test the entry point used in our plugin implementations. And we don't even care if anything else in those interfaces/classes changes because the only thing that is releveant to us is (in most cases) the TRUE/FALSE returned from the one, single method that we invoke on these external services.

> It still appears that the mocks will let us build plug-ins, and "pass" them, when the plugins will not actual work when integrated.

This can only ever happen if we a) write wrong tests or b) the other system is broken (which is nothing we can account for in all the cases anyways). We have to trust that the services that we are using are functional in a fully bootstrapped Drupal. Our only worry should be (in the plugin testing scope) to test whether our implementation of the plugin is correct and whether we are invoking the right methods on the right services at the right time and the plugin returns the expected values when done so.

PR #64, opened 2014-06-13
Context passing between Rules components

klausi:
Issue https://drupal.org/node/2245039

### DIFF
--- a/src/Engine/RulesConditionBase.php
+++ b/src/Engine/RulesConditionBase.php
@@ -116,4 +116,11 @@ abstract class RulesConditionBase extends ContextAwarePluginBase implements Rule
     return [];
   }
 
+  /**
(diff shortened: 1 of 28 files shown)


src/Engine/RulesConditionBase.php
```diff
@@ -116,4 +116,11 @@ public function calculateDependencies() {
     return [];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return '';
```
fubhy: Why do we need this? Isn't it kinda a requirement of a plugin to define this in any case?

klausi: I added it because I was too lazy to implement it in the the test plugins, but yeah, this should probably just be moved to the test plugins.

PR #65, opened 2014-06-14
Refactoring context system overrides out of Rules.

fubhy:


fubhy:
This is based on https://drupal.org/node/2281635 and will obviously fail at the moment because that Patch is not in core yet. It works when the patch is applied though. I didn't want to change our .travis.yml though :P

fubhy:
Updated the patch to reflect the latest changes in the patch in https://drupal.org/node/2281635

fubhy:
This is now green (locally and with the patch from https://drupal.org/node/2281635 applied to Core).

PR #68, opened 2014-06-19
Fix invocation of phpunit in Travis CI test suite.

klausi:
After the core phpunit 4.x update it seems that the phpunit tests do not get executed anymore. See https://travis-ci.org/fago/rules/jobs/27966468 for example.

This pull request should fail the Travis check because it contains a PHP syntax error.


klausi:
Hooray, it fails! Ok, so we cannot use the symlink in the bin directory anymore (although it still works on my local machine).

PR #66, opened 2014-06-18
- Initial attempt at implementing the data value is empty condition.

jantoine1:
Hey Fago,

It was great working with you on Rule at the DrupalCon Austin sprint. I want to get more involved with the Rules team, but I need to clear some client projects off my plate first. In the mean time, I didn't want to hold up https://drupal.org/node/2281081 and also didn't want my work to go to waste, so it is contained within this pull request. Let me know if you have any questions.

Cheers,

jantoine

PR #69, opened 2014-06-24
2281081 - Port "Data value is empty (data_is_empty)" condition to 8.x

paranojik:
This is basically a copy of https://github.com/fago/rules/pull/66 with added unit test.
Related d.g.o. issue: https://www.drupal.org/node/2281081.

### DIFF
--- a/tests/src/Condition/DataIsEmptyTest.php
+++ b/tests/src/Condition/DataIsEmptyTest.php
@@ -1,0 +1,122 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Tests\Condition\DataIsEmptyTest.
+ */
+
+namespace Drupal\rules\Tests\Condition;
+
+use Drupal\rules\Plugin\Condition\DataIsEmpty;
+use Drupal\Core\Plugin\Context\ContextDefinition;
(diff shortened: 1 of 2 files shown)


tests/src/Condition/DataIsEmptyTest.php
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Tests context value setting and getting.
+   *
+   * @covers ::setContextValue()
+   * @covers ::getContextValue()
+   */
+  public function testContextValue() {
+    // We can't mock the NodeInterface because there is a bug in PHPUnit below
+    // version 3.8 that causes mocking of interfaces that extend \Traversable
+    // to fail. @see https://github.com/sebastianbergmann/phpunit-mock-objects/issues/103
```
klausi: Drupal core now uses PHPunit 4, so we should be able to use the interface now. Please check.

paranojik:
Yup. Works just fine. Thanks, @klausi!

klausi:
The context stuff just landed in core and @fubhy updated all the condition plugins, so you will have to update the annotation, too (and remove the static context method).

paranojik:
Context stuff updated...

PR #70, opened 2014-06-24
Move documentation to separate repository.

fubhy:
I set up the documentation on gitbook.io: http://thefubhy.gitbooks.io/rules

The repository that drives the book is at https://github.com/fubhy/rules-docs

I think it makes perfect sense to separate the documentation (which might is hopefully going to grow quickly) to a separate repository to keep our commit log clean from documentation changes, typo fixes, etc. and to reduce the overall size of the repository and keep it clean.

We can move the rules-docs repo to fago's github account later and also set it up to be hosted on his gitbooks.io account once he has one if that is desirable.

I think we should also keep maintaining the documentation on github/gitbooks as it gives us a beautiful and nicely usable, feature-rich stack for hosting the documentation. It also distributes the docs as pdf, epub, etc. as downloads and there is the possibility to add disqus commenting functionality if we want that.


klausi:
I think we should keep the README files as is, since we are already sending people here to lookup how to get started on development. It is the most important minimum documentation that should stay in the README, right?

fubhy:
I'd argue that as soon as we got images, etc. in place that the README is too cluttered for the main project README and should be simplified and moved somewhere else. The documentation is important, especially the contribution part of course but a the main project README should just be a short introduction including links for further information imho.

klausi:
OK, so let's just move the environment section with the screenshots from the README and leave the rest as is?

PR #76, opened 2014-07-02
Add context/provided context handling for Rules actions

klausi:
See also: https://www.drupal.org/node/2245045


klausi:
Test cases for the actions consuming/providing context is missing right now, same as we have for conditions.

PR #78, opened 2014-07-02
Adding 'save entity' action.

fubhy:


src/Plugin/Action/SaveEntity.php
```diff
@@ -0,0 +1,56 @@
...
+   */
+  public function summary() {
+    return $this->t('Save entity.');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    // @todo Properly implement this.
```
klausi: This todo should say: todo mark the entity for saving later once that feature is implemented in the Rules state.

PR #79, opened 2014-07-02
Adding 'create any path alias' action.

fubhy:


src/Plugin/Action/CreatePathAlias.php
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $source = $this->getContextValue('source');
+    $alias = $this->getContextValue('alias');
+    $language = $this->getContextValue('language') ?: LanguageInterface::LANGCODE_NOT_SPECIFIED;
+    $this->aliasStorage->save($source, $alias, $language);
```
klausi: That looks wrong. $language can now be an object or a string? But AliasStorageInterface expects a string for $langcode? Anything you pass to the alias storage should be named $langcode, to make it clear that it is not an object.

PR #59, opened 2014-06-08
Initial commit for data comparison condition and associated test

diarmy:
I've added a @todo for getContextDefinitions()

PR #84, opened 2014-07-10
2281079 - Data comparison (data_is) condition implementation.

paranojik:
Initial commit for data comparison condition.

Based on https://github.com/fago/rules/pull/59.

### DIFF
--- a/src/Plugin/Condition/DataComparison.php
+++ b/src/Plugin/Condition/DataComparison.php
@@ -20,7 +20,7 @@ use Drupal\rules\Engine\RulesConditionBase;
  *       label = @Translation("Data to compare"),
  *       description = @Translation("The data to be checked to be empty, specified by using a data selector, e.g. 'node:uid:entity:name:value'.")
  *     ),
- *     "op" = @ContextDefinition("string",
+ *     "operator" = @ContextDefinition("string",
  *       label = @Translation("Operator"),
  *       description = @Translation("The comparison operator."),
  *       required = FALSE
@@ -50,10 +50,10 @@ class DataComparison extends RulesConditionBase {
    */
   public function evaluate() {
     $data = $this->getContextValue('data');
-    $op = $this->getContext('op')->getContextData() ? $this->getContextValue('op') : '==';
+    $operator = $this->getContext('operator')->getContextData() ? $this->getContextValue('operator') : '==';
     $value = $this->getContextValue('value');
     
-    switch ($op) {
+    switch ($operator) {
       case '<':
         return $data < $value;
 
@@ -66,7 +66,6 @@ class DataComparison extends RulesConditionBase {
       case 'IN':
         return is_array($value) && in_array($data, $value);
 
-      case '==':
       default:
         // In case both values evaluate to FALSE, further differentiate between
         // NULL values and values evaluating to FALSE.
--- a/src/Plugin/Condition/DataComparison.php
+++ b/src/Plugin/Condition/DataComparison.php
@@ -1,0 +1,117 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Condition\DataComparison.
+ */
+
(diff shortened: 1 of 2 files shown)


src/Plugin/Condition/DataComparison.php
```diff
@@ -0,0 +1,117 @@
...
+      case '==':
+      default:
+        return $this->equals($data, $value);
+    }
+  }
+
+  /**
+   * Evaluates the "isLessThan" condition.
+   */
+  protected function isLessThan($data, $value) {
```
klausi: It does not make sense now to have that method, because you are not unit testing it? Since this is just one line and the evaluate() method is not that complicated either I would suggest to just directly invoke the operations in the evaluate() method. I know @fubhy suggested it, but let's not overcomplicate things.

fubhy: I actually only suggested that to make it easier in the unit test, yeah...

tests/src/Condition/DataComparisonTest.php
```diff
@@ -0,0 +1,200 @@
...
+      ->setContextValue('data', $this->getMockTypedData(['Kitten']))
+      ->setContextValue('op', $this->getMockTypedData('contains'))
+      ->setContextValue('value', $this->getMockTypedData(['Llama']));
+    $this->assertFalse($this->condition->evaluate());
+  }
+
+  /**
+   * Tests evaluating the condition with the "IN" operator.
+   */
+  public function testConditionEvaluationOperatorIN() {
```
fubhy: Should be "In" not IN

tests/src/Condition/DataComparisonTest.php
```diff
@@ -0,0 +1,200 @@
...
+   * @covers ::summary()
+   */
+  public function testSummary() {
+    $this->assertEquals('Data comparison', $this->condition->summary());
+  }
+
+  /**
+   * Tests evaluating the condition with the "equals" operator.
+   */
+  public function testConditionEvaluationOperatorEquals() {
```
fubhy, line 69: Your test cases lack the "@covers" documentation.

paranojik:
Not sure if this is right but I went both ways... kept the evaluate() in one piece and the unit test as separate methods for each operator.

src/Plugin/Condition/DataComparison.php
```diff
@@ -0,0 +1,80 @@
...
+      case '>':
+        return $data > $value;
+
+      case 'contains':
+        return is_string($data) && strpos($data, $value) !== FALSE || is_array($data) && in_array($value, $data);
+
+      case 'IN':
+        return is_array($value) && in_array($data, $value);
+
+      case '==':
```
fubhy: That line is redundant if you want to go with "default".

I am actually not sure if "default" is good. Imho it should fail if you have a non-supported operator.

src/Plugin/Condition/DataComparison.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @Condition(
+ *   id = "rules_data_is",
+ *   label = @Translation("Data comparison"),
+ *   context = {
+ *     "data" = @ContextDefinition("any",
+ *       label = @Translation("Data to compare"),
+ *       description = @Translation("The data to be checked to be empty, specified by using a data selector, e.g. 'node:uid:entity:name:value'.")
+ *     ),
+ *     "op" = @ContextDefinition("string",
```
fubhy: We just recently agreed that we don't want to use any abbreviations anymore (not for contexts, and not for variables either). So "op" should be "operator" and $op should be $operator.

paranojik:
...like so...?

fubhy:
Yep. Pinging @fago to check on the 'default' switch. Wondering if we should throw an exception instead? You know... Instead of silently doing a potentially wrong comparison if somehow an unsupported operator was specified.

PR #86, opened 2014-07-12
Replace getInfo() in tests with native phpDoc +...

jibran:
... annotations (following PHPUnit).
After https://www.drupal.org/node/2301125


jibran:
Moved to #87

PR #88, opened 2014-07-13
Added PHPCS to .travis.yml.

jibran:

PR #89, opened 2014-07-13
Fixed PHPCS errors.

jibran:
We can test it after #88.


src/Plugin/Condition/DataComparison.php
```diff
@@ -49,31 +49,37 @@ public function summary() {
...
     $data = $this->getContextValue('data');
     $operator = $this->getContext('operator')->getContextData() ? $this->getContextValue('operator') : '==';
     $value = $this->getContextValue('value');
-    
+
     switch ($operator) {
       case '<':
-        return $data < $value;
+        $result = $data < $value;
+        break;
```
klausi: Why do we need this change? This is fine as is?

jibran: Please see https://travis-ci.org/fago/rules/jobs/29828127#L179

klausi: Ah, so there is a bug in Coder? There is a newline after the case breaking return statement, so this is fine?

jibran: Yeah this is bug. But I couldn't fix it till I add break and I can't add break after return so this is the only way to fix it.

klausi: So let's not "fix" stuff that is not broken. I fear that PHPCS for Drupal is not quite as mature as we would like it to be right now. Let's file bug reports against Coder instead.

klausi:
Squashed and merged, thanks!

PR #95, opened 2014-08-07
add variable action

dawehner:
not done yet

### DIFF
--- a/src/Plugin/Action/AddVariable.php
+++ b/src/Plugin/Action/AddVariable.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Action\AddVariable.
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Engine\RulesActionBase;
+
+/**
+ * @Action(
+ *   id = "rules_add_variable",
+ *   label = @Translation("Add a variable"),
+ *   context = {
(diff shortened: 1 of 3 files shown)


fago:
please also create a d.o. issue and reference it at https://www.drupal.org/node/2269577

src/Plugin/Action/AddVariable.php
```diff
@@ -0,0 +1,38 @@
...
+ * )
+ */
+class AddVariable extends RulesActionBase {
+
+  /**
+   * Executes the plugin.
+   */
+  public function execute() {
+    $this->setProvidedValue('variable_added', $this->getContext('value'));
+    return TRUE;
```
klausi, line 35: This method should not return anything.

src/Tests/AddVariableIntegrationTest.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Drupal\rules\Tests;
+
+/**
+ * Test the data processor plugins during Rules evaluation.
+ *
+ * @group rules
+ */
+class AddVariableIntegrationTest extends RulesDrupalTestBase {
```
klausi, line 15: Why do we need a Simpletest for this and cannot use the PHPUnit test base? Please add a comment.

dasjo:
this is the according issue https://www.drupal.org/node/2409049

dasjo:
see the follow-up https://github.com/fago/rules/pull/121

PR #98, opened 2014-08-07
Added remove item from list action.

foobarna:
Issue 2317227 in drupal.org

### DIFF
--- a/src/Plugin/Action/RemoveListItem.php
+++ b/src/Plugin/Action/RemoveListItem.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Action\RemoveListItem.
+ */
+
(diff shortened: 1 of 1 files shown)


src/Plugin/Action/RemoveListItem.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Provides a 'Remove item from list' action.
+ *
+ * @Action(
+ *   id = "rules_list_item_remove",
+ *   label = @Translation("Remove item from list"),
+ *   context = {
+ *    "list" = @ContextDefinition("list",
+ *      label = @Translation("List"),
+ *      description = @Translation("The data list for which an item is to be removed.")
```
fago, line 19: You are missing @todos for all the not yet supported and here missing keys from the d7 version.

klausi:
This continues at #116

PR #99, opened 2014-08-08
Initial port of ListContains condition to 8.x

stephenpurkiss:


src/Plugin/Condition/ListContains.php
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $list = $this->getContextValue('list');
+    $item = $this->getContextValue('item');
+
+    if ($item instanceof EntityInterface && $id = $item->id()) {
+      // Check for equal items using the identifier if there is one.
+      foreach ($list->getValue() as $i) {
```
fago: always use verbose, readable variable names -> $item or $list_item

tests/src/Condition/ListContainsTest.php
```diff
@@ -0,0 +1,67 @@
...
+    $this->condition
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Two');
+    $this->assertTrue($this->condition->evaluate());
+
+    // Test that the list doesn't contain 'Five'.
+    $this->condition
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Five');
+    $this->assertFalse($this->condition->evaluate());
```
fago, line 140: misses test coverage for entities, however I'm not sure it will have to work the same way as in d7 for entities.

PR #102, opened 2014-08-09
#2317193 | Port "Add an item to a list" action to D8

stephenpurkiss:
Initial port

### DIFF
--- a/src/Plugin/Action/DataListItemAdd.php
+++ b/src/Plugin/Action/DataListItemAdd.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Action\DataListItemAdd.
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Engine\RulesActionBase;
+
+/**
+ * Provides an 'Add list item' action.
+ *
+ * @Action(
+ *   id = "rules_list_item_add",
+ *   label = @Translation("Add list item"),
+ *   context = {
+ *     "list" = @ContextDefinition("list",
+ *       label = @Translation("List"),
+ *       description = @Translation("The data list, to which an item is to be added.")
+ *     ),
+ *     "item" = @ContextDefinition("any",
+ *       label = @Translation("Item"),
+ *       description = @Translation("Item to add.")
+ *     ),
+ *     "unique" = @ContextDefinition("boolean",
+ *       label = @Translation("Enforce uniqueness"),
+ *       description = @Translation("Only add the item to the list if it is not yet contained."),
+ *       required = FALSE
+ *     ),
+ *     "pos" = @ContextDefinition("string",
+ *       label = @Translation("Insert position"),
+ *       description = @Translation("Position to insert the item."),
+ *       required = FALSE
+ *     )
+ *   },
+ *   provides = {
+ *     "outputlist" = @ContextDefinition("list",
+ *       label = @Translation("The resulting data list with item added.")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7?
+ * @todo: Add group information from Drupal 7?
+ * @todo: set ContextDefinition restriction
+ * @todo: Add info alter
+ */
+class DataListItemAdd extends RulesActionBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t('Add list item');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $list = $this->getContextValue('list');
+    $item = $this->getContextValue('item');
+    $position = ($this->getContextValue('pos') ? $this->getContextValue('pos') : 'end');
+    $unique = ($this->getContextValue('unique') ? $this->getContextValue('unique') : FALSE);
+    // Optionally, only add the list item if it is not yet contained.
+    if (!((bool) $unique && in_array($item, $list))) {
+      if ($position === 'start' ) {
+        array_unshift($list, $item);
+      }
+      else {
+        $list[] = $item;
+      }
+    }
+    $this->setContextValue('list', $list);
+  }
+}
--- a/tests/src/Integration/Action/DataListItemAddTest.php
+++ b/tests/src/Integration/Action/DataListItemAddTest.php
@@ -1,0 +1,162 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Tests\rules\Integration\Action\DataListItemAddTest.
+ */
+
+namespace Drupal\Tests\rules\Integration\Action;
+
+use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
+
+/**
+ * @coversDefaultClass \Drupal\rules\Plugin\Action\DataListItemAdd
+ * @group rules_actions
+ */
+class DataListItemAddTest extends RulesIntegrationTestBase  {
+
+  /**
+   * The action to be tested.
+   *
+   * @var \Drupal\rules\Engine\RulesActionInterface
+   */
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
+    $this->action = $this->actionManager->createInstance('rules_list_item_add');
+  }
+
+  /**
+   * Tests the summary.
+   *
+   * @covers ::summary()
+   */
+  public function testSummary() {
+    $this->assertEquals('Add list item', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution with default values.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecutionWithDefaults() {
+    // Test adding a value at the end (default position).
+    $list = ['One', 'Two', 'Three'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four');
+
+    $this->action->execute();
+
+    // The list should contain four items, with the new item at the end.
+    $this->assertArrayEquals([
+        'One',
+        'Two',
+        'Three',
+        'Four'
+      ], $this->action->getContextValue('list'));
+  }
+
+  /**
+   * Tests the action execution - item append.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecutionItemAppend() {
+    // Test adding a value at the end.
+    $list = ['One', 'Two', 'Three'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four')
+      ->setContextValue('pos', 'end');
+
+    $this->action->execute();
+
+    // The list should contain four items, with the new item added at the end.
+    $this->assertArrayEquals([
+        'One',
+        'Two',
+        'Three',
+        'Four'
+      ], $this->action->getContextValue('list'));
+  }
+
+  /**
+   * Tests the action execution - item prepend.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecutionItemPrepend() {
+    // Test adding a value at the start.
+    $list = ['One', 'Two', 'Three'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Zero')
+      ->setContextValue('pos', 'start');
+
+    $this->action->execute();
+
+    // The list should contain four items, with the new item added at the start.
+    $this->assertArrayEquals([
+        'Zero',
+        'One',
+        'Two',
+        'Three'
+      ], $this->action->getContextValue('list'));
+  }
+
+  /**
+   * Tests the action execution - enforce unique items.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecutionEnforceUnique() {
+    // Test unique.
+    $list = ['One', 'Two', 'Three', 'Four'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four')
+      ->setContextValue('unique', TRUE);
+
+    $this->action->execute();
+
+    // The list should remain the same.
+    $this->assertArrayEquals([
+        'One',
+        'Two',
+        'Three',
+        'Four'
+      ], $this->action->getContextValue('list'));
+  }
+
+  /**
+   * Tests the action execution - add non-unique items.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecutionNonUnique() {
+    // Test non-unique.
+    $list = ['One', 'Two', 'Three', 'Four'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four')
+      ->setContextValue('unique', FALSE)
+      ->setContextValue('pos', 'end');
+
+    $this->action->execute();
+
+    // The list should contain five items, with the new item added at the end.
+    $this->assertArrayEquals(['One', 'Two', 'Three', 'Four', 'Four'], $this->action->getContextValue('list'));
+  }
+}


tests/src/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Tests the action execution.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
+
+    // Test adding to array of string values
+    $list = array('One','Two','Three');
```
fago: misse spaces between the items; ie.after the comma

tests/src/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    // Test that the list contains 'Four'.
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four');
+    $this->assertArrayEquals(array('list' => array('One','Two','Three','Four')), $this->action->execute());
```
fago: spaces after commas again.

src/Plugin/Action/ListAddItem.php
```diff
@@ -0,0 +1,84 @@
...
+
+      case 'end':
+        $list[] = $item;
+        break;
+
+      default:
+        $list[] = $item;
+        break;
+    }
+    return array('list' => $list);
```
fago: Providing context value works now differently in d8. Check other actions which do it already, e.g. adding a variable (it's in an issue) or the test-coverage for it.

stephenpurkiss:
Sorted (I think!) the spacing issue, having issues working out the git branch/merging thing hence the references to temp branch etc. Looking into context value issue next.

src/Plugin/Action/ListAddItem.php
stephenpurkiss: Could you elaborate? I'm unsure as to what's the issue here - I am getting those values from the context in the lines just above what you're highlighting:

```
$list = $this->getContextValue('list');
$item = $this->getContextValue('item');
$position = $this->getContextValue('pos');
```

stephenpurkiss: ...perhaps you mean the return value needs to be set as context, i.e. something like this:

Instead of:
return array('list' => $list);

it should be
$this->setContextValue('list', $list);

I'll keep hunting...

klausi: $this->setProvidedValue(..), example: https://github.com/fago/rules/blob/8.x-3.x/tests/modules/rules_test/src/Plugin/Action/TestStringAction.php

klausi:
Do you want to continue working on this, Steve? Any help needed?

stephenpurkiss:
hmm I'd completely forgotten about this, sorry! Sure, I think it may just need squashing but will check

stephenpurkiss:
Oh hang on, forgot about the unit test changes. brb

stephenpurkiss:
Oh, OK, it's a bit further behind than I thought - pointers welcome if you have time and I'm not slowing the process down too much. I've got to figure out how to set up a test for a list, code doesn't seem to be there at the moment (my understanding has increased in the last 22 days!).

Will take a look over the weekend, and I'll be in Amsterdam this time next week so around for the first lot of extended sprints but then flying back the next friday - no doubt say hi if you're going too - not long now yay!

tests/src/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Tests\Action\ListAddItemTest.
+ */
+
+namespace Drupal\rules\Tests\Action;
```
klausi: Namespace should be moved to the "Unit" subfolder, please rebase your work on the 8.x-3.x branch and move the test.

stephenpurkiss:
Managed a bit of time on it, now stuck at "PHP Fatal error: Call to a member function get() on a non-object in /Users/steve/Sites/d8dev/core/lib/Drupal.php on line 428" when I run the tests locally :(

Will continue

tests/src/Unit/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,68 @@
...
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
+
+    // Test adding to array of string values
+    $list = array('One', 'Two', 'Three');
+
+    // Test that the list contains 'Four'.
+    $this->action
+      ->setContextValue('list', $list)
```
klausi: I think you need to use $this->getMockTypedData($list) here, take a look at the other unit tests. Maybe also $this->action->setTypedDataManager($this->getMockTypedDataManager()); in setUp().

tests/src/Unit/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\Tests\rules\Unit\RulesUnitTestBase;
+
+/**
+ * @coversDefaultClass \Drupal\rules\Plugin\Action\ListAddItem
+ * @group rules_action
+ *
+ * @todo: add more tests including coverage for unique and position
+ */
+class ListAddItemTest extends RulesUnitTestBase {
```
fago: Use RulesIntegrationTestBase now - we've got a new test base.

tests/src/Unit/Action/ListAddItemTest.php
fago: Nope, let's use RulesIntegrationTestBase now and the real typed data manager. So we can save converting this test later on :)

stephenpurkiss:
Hi - am stuck on two issue on this one, just spoke to fago who didn't know what the issue is either so posting here whilst trying to debug too, although may move to something else whilst I'm stuck!

I'm setting a provided value but it says the context is not valid. Also I'm having to force values for 'unique' and 'pos' even though I say they're not required. Here's the output of my test:

There was 1 error:

1) Drupal\Tests\rules\Unit\Action\ListAddItemTest::testActionExecution
Drupal\Component\Plugin\Exception\ContextException: The outputlist provided context is not valid.

/Users/steve/Sites/d8dev/modules/rules/src/Context/RulesContextTrait.php:61
/Users/steve/Sites/d8dev/modules/rules/src/Context/RulesContextTrait.php:50
/Users/steve/Sites/d8dev/modules/rules/src/Context/RulesContextTrait.php:40
/Users/steve/Sites/d8dev/modules/rules/src/Plugin/Action/ListAddItem.php:90
/Users/steve/Sites/d8dev/modules/rules/tests/src/Unit/Action/ListAddItemTest.php:66
phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:179
phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:132

## 

There was 1 failure:

1) Drupal\Tests\rules\Unit\Condition\UserHasEntityFieldAccessTest::testConditionEvaluation
Failed asserting that false is true.

/Users/steve/Sites/d8dev/modules/rules/tests/src/Unit/Condition/UserHasEntityFieldAccessTest.php:123
phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:179
phar:///usr/local/bin/phpunit/phpunit/TextUI/Command.php:132

FAILURES!

klausi:
The UserHasEntityFieldAccessTest.php should be fixed in 8.x-3.x, please merge your branch.

tests/src/Unit/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
+    $this->action = new ListAddItem([], '', ['context' => [
+      'list' => new ContextDefinition('list'),
+      'item' => new ContextDefinition('string')
+    ]]);
```
klausi: The outputlist provided context is missing here? Same for the other contexts in the annotation?

stephenpurkiss: > The UserHasEntityFieldAccessTest.php should be fixed in 8.x-3.x, please merge your branch.

Ah - I think the issue here is I fetched the upstream and rebased to my local 8.x-3.x but of course this is a branch so I'll need to now rebase that branch on the latest 8.x-3.x.

> The outputlist provided context is missing here? Same for the other contexts in the annotation?

The 'pos' and 'unique' contexts aren't required - I take it then I still have to declare them here. Thanks for the pointer.

stephenpurkiss:
So I've updated my branch, added values for 'unique' and 'pos', but still getting provided context is not valid :(

stephenpurkiss:
I'm just print_r'ing out the input and output to ensure they are arrays, seems to be all ok in that respect:

..............EinputArray
(
    [0] => One
    [1] => Two
    [2] => Three
)
resultArray
(
    [0] => One
    [1] => Two
    [2] => Three
    [3] => Four
)
.................................................. 65 / 92 ( 70%)
...........................

Time: 664 ms, Memory: 21.00Mb

stephenpurkiss:
So through debugging I can see it's just not returning a definition of the context 'provides' in the getProvidedDefinition function, it's just blank. My declaration seems ok so not sure what the issue is, keeping hunting...

tests/src/Unit/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,72 @@
...
+   */
+  public function setUp() {
+    parent::setUp();
+
+    $this->action = new ListAddItem([], '', ['context' => [
+      'list' => new ContextDefinition('list'),
+      'item' => new ContextDefinition('string'),
+      'unique' => new ContextDefinition('boolean'),
+      'pos' => new ContextDefinition('string')
+    ]]);
```
klausi: 'provides' is missing here.

stephenpurkiss: ah ok cool thx

stephenpurkiss:
Hmmm... OK so I fixed that last issue and managed to make it pass the tests but only by returning a value as well as setting the provided value:

```
$this->setProvidedValue('outputlist', $list);
return array('list' => $list);
```

stephenpurkiss:
...which makes me think I don't need the setProvidedValue line then, I'll take that out & resubmit

stephenpurkiss:
OK, so that worked ;)

stephenpurkiss:
Forgot to remove my forcing values test code, removed now.

stephenpurkiss:
Added two more tests, one to check 'position' option works, one to check 'unique' option works.

stephenpurkiss:
I put all three tests in testActionExecution - not sure if they should be split up to a function for each? I see it kinda auto-finds tests as long as I have the @covers::execute there so presume I can do things like testActionExecutionWithPosition, testActionExecutionUnique, etc.?

stephenpurkiss:
Oh hang on, wrote a test for unique but not not checking unique. On way back to hotel so will write it there. 

There's always something... ;)

stephenpurkiss:
Non-unique test added.

src/Plugin/Action/ListAddItem.php
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Engine\RulesActionBase;
+
+/**
+ * Provides an 'Add a list item' action.
+ *
+ * @Action(
+ *   id = "rules_action_data_list_add",
```
fago: Should be rules_list_add.

src/Plugin/Action/ListAddItem.php
```diff
@@ -0,0 +1,88 @@
...
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7?
+ * @todo: Add group information from Drupal 7?
+ * @todo: set ContextDefinition restriction
+ * @todo: Add info alter
+ * @todo: Add code to deal with default value for position
+ * @todo: Replace in_array() with \Drupal\rules\Plugin\Condition\ListContains
+ */
+class ListAddItem extends RulesActionBase {
```
fago: Should be DataListAddItem

tests/src/Unit/Action/ListAddItemTest.php
```diff
@@ -0,0 +1,107 @@
...
+   */
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
+    $this->action = new ListAddItem([], '', ['context' => [
```
fago: This should use the actionmanager to instantiate the plugin.

src/Plugin/Action/ListAddItem.php
stephenpurkiss: ok

src/Plugin/Action/ListAddItem.php
stephenpurkiss: Presuming DataListItemAdd as there's now a DataListItemRemove and we're naming: ThingAction

src/Plugin/Action/ListAddItem.php
stephenpurkiss: changing to rules_list_item_add to keep in toe with rules_list_item_remove which has already been committed to 8.x-3.x

tests/src/Unit/Action/ListAddItemTest.php
stephenpurkiss: OK, am also seeing DataListItemRemoveTest.php is in Drupal\Tests\rules\Integration\Action so moving this over there too.

stephenpurkiss:
Wondering if this is still wanted to be done:
- @todo: Replace in_array() with \Drupal\rules\Plugin\Condition\ListContains

src/Plugin/Action/DataListItemAdd.php
```diff
@@ -0,0 +1,85 @@
...
+    $position = ($this->getContextValue('pos') ? $this->getContextValue('pos') : 'end');
+    $unique = ($this->getContextValue('unique') ? $this->getContextValue('unique') : FALSE);
+    // Optionally, only add the list item if it is not yet contained.
+    if (!((bool) $unique && in_array($item, $list))) {
+      switch ($position) {
+        case 'start':
+          array_unshift($list, $item);
+          break;
+
+        case 'end':
```
klausi: So the default branch and the "end" branch do the same? Then we can just use "default" alone.

tests/src/Integration/Action/DataListItemAddTest.php
```diff
@@ -0,0 +1,115 @@
...
+      ->setContextValue('item', 'Four')
+      ->setContextValue('unique', FALSE)
+      ->setContextValue('pos', 'end');
+
+    $this->action->execute();
+
+    // The list should contain five items, with the new item added at the end.
+    $this->assertArrayEquals(['One', 'Two', 'Three', 'Four', 'Four'], array_values($this->action->getContextValue('list')));
+  }
+}
```
klausi: Missing newline at the end.

tests/src/Integration/Action/DataListItemAddTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $list = ['One', 'Two', 'Three'];
+
+    $this->action
+      ->setContextValue('list', $list)
+      ->setContextValue('item', 'Four');
+
+    $this->action->execute();
+
+    // The list should contain four items, with the new item at the end.
+    $this->assertArrayEquals(['One', 'Two', 'Three', 'Four'], array_values($this->action->getContextValue('list')));
```
klausi: why is the array_values() necessary here? It should also work without it?

tests/src/Integration/Action/DataListItemAddTest.php
```diff
@@ -0,0 +1,115 @@
...
+  public function testSummary() {
+    $this->assertEquals('Add list item', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
```
klausi: I think we should split this up to multiple test methods, one for each scenario.

src/Plugin/Action/DataListItemAdd.php
```diff
@@ -0,0 +1,85 @@
...
+ *       label = @Translation("The resulting data list with item added.")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7?
+ * @todo: Add group information from Drupal 7?
+ * @todo: set ContextDefinition restriction
+ * @todo: Add info alter
+ * @todo: Replace in_array() with \Drupal\rules\Plugin\Condition\ListContains
```
klausi: Yes, I agree that this todo can be removed.

src/Plugin/Action/DataListItemAdd.php
stephenpurkiss: Changed to an if..else

tests/src/Integration/Action/DataListItemAddTest.php
stephenpurkiss: fixed

tests/src/Integration/Action/DataListItemAddTest.php
stephenpurkiss: just copied from previous code somewhere. I see DataListItemRemove used this too and got through ok so presumed was ok, hadn't really thought about it but yes, does seem unused so have removed.

tests/src/Integration/Action/DataListItemAddTest.php
stephenpurkiss: Indeed, good point. Looked quickly at different tests, didn't seem to be any particular naming pattern so just chose relevant. I'll RTFM at some point... ;)

src/Plugin/Action/DataListItemAdd.php
stephenpurkiss: ok done. I've squashed the commits so it's ready but guess really I shouldn't do that until you request otherwise makes it harder to see the incremental changes. I'll try to be less squash-trigger-happy ;)

tests/src/Integration/Action/DataListItemAddTest.php
stephenpurkiss: Wondering if it has anything to do with array value injection security:

https://twitter.com/skwashd/status/522415328119107584

http://cgit.drupalcode.org/drupal/commit/?id=26a7752

diff --git a/includes/database/database.inc b/includes/database/database.inc
index f78098b..01b6385 100644
--- a/includes/database/database.inc
+++ b/includes/database/database.inc
@@ -736,7 +736,7 @@ abstract class DatabaseConnection extends PDO {
     // to expand it out into a comma-delimited set of placeholders.
     foreach (array_filter($args, 'is_array') as $key => $data) {
       $new_keys = array();
-      foreach ($data as $i => $value) {
- ```
   foreach (array_values($data) as $i => $value) {
   // This assumes that there are no other placeholders that use the same
   // name.  For example, if the array placeholder is defined as :example
   // and there is already an :example_2 placeholder, this will generate
  ```
-  /**
-   \* Test SQL injection via database query array arguments.
-   */
-  public function testArrayArgumentsSQLInjection() {
-    // Attempt SQL injection and verify that it does not work.
-    $condition = array(
-      "1 ;INSERT INTO {test} SET name = 'test12345678'; -- " => '',
-      '1' => '',
-    );
-    try {
-      db_query("SELECT \* FROM {test} WHERE name = :name", array(':name' => $condition))->fetchObject();
-      $this->fail('SQL injection attempt via array arguments should result in a PDOException.');
-    }
-    catch (PDOException $e) {
-      $this->pass('SQL injection attempt via array arguments should result in a PDOException.');
-    }
  +
-    // Test that the insert query that was used in the SQL injection attempt did
-    // not result in a row being inserted in the database.
-    $result = db_select('test')
-      ->condition('name', 'test12345678')
-      ->countQuery()
-      ->execute()
-      ->fetchField();
-    $this->assertFalse($result, 'SQL injection attempt did not result in a row being inserted in the database table.');
-  }
  +
  }
  
  /**

stephenpurkiss: Ugh, sorry that paste didn't work out well. Key point:
-      foreach ($data as $i => $value) {
-      foreach (array_values($data) as $i => $value) {

PR #105, opened 2014-08-15
Initial port of PathAliasDelete with test. Currently works for deleting ...

stephenpurkiss:
...a single url alias.

The original D7 function was the create or delete URL alias and thus had two inputs - source and alias. If the source was missing but an alias provided it would delete the alias, and if the alias was missing and a source provided then it would delete aliases associated with that source. 

Now that the create and delete functions are split, the delete only needs one input, i.e. what you want to delete. This initial patch provides the code to delete a single url alias as the class name suggests. 

I believe further discussion is needed to decide how this functionality is going to work in 8, i.e. can we still pass it a source and delete all relevant aliases (how do we check whether it's source or alias?) - can we send it an array of sources and/or aliases, etc. 

So, for the moment, this is a start.

### DIFF
--- a/src/Plugin/Action/AliasDelete.php
+++ b/src/Plugin/Action/AliasDelete.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Action\AliasDelete.
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\Core\Path\AliasStorageInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Engine\RulesActionBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Delete any path alias' action.
+ *
+ * @Action(
+ *   id = "rules_path_alias_delete",
+ *   label = @Translation("Delete any path alias"),
+ *   context = {
+ *     "alias" = @ContextDefinition("string",
+ *       label = @Translation("Existing system path alias"),
+ *       description = @Translation("Specifies the existing path alias you wish to delete, for example 'about/team'. Use a relative path and do not add a trailing slash.")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add group information from Drupal 7.
+ */
+class AliasDelete extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The alias storage service.
+   *
+   * @var \Drupal\Core\Path\AliasStorageInterface
+   */
+  protected $aliasStorage;
+
+  /**
+   * Constructs a AliasDelete object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Path\AliasStorageInterface $alias_storage
+   *   The alias storage service.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, AliasStorageInterface $alias_storage) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->aliasStorage = $alias_storage;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('path.alias_storage')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t('Delete any path alias');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $alias = $this->getContextValue('alias');
+    $this->aliasStorage->delete(['alias' => $alias]);
+  }
+}
--- a/tests/src/Unit/Action/AliasDeleteTest.php
+++ b/tests/src/Unit/Action/AliasDeleteTest.php
@@ -1,0 +1,77 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Tests\rules\Unit\Action\AliasDeleteTest.
+ */
+
+namespace Drupal\Tests\rules\Unit\Action;
+
+use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\rules\Plugin\Action\AliasDelete;
+use Drupal\Tests\rules\Unit\RulesUnitTestBase;
+
+/**
+ * @coversDefaultClass \Drupal\rules\Plugin\Action\AliasDelete
+ * @group rules_action
+ */
+class AliasDeleteTest extends RulesUnitTestBase {
+
+  /**
+   * The action to be tested.
+   *
+   * @var \Drupal\rules\Engine\RulesActionInterface
+   */
+  protected $action;
+
+  /**
+   * The mocked alias storage service.
+   *
+   * @var \PHPUnit_Framework_MockObject_MockObject|\Drupal\Core\Path\AliasStorageInterface
+   */
+  protected $aliasStorage;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
+    $this->aliasStorage = $this->getMock('Drupal\Core\Path\AliasStorageInterface');
+
+    $this->action = new AliasDelete([], '', ['context' => [
+      'alias' => new ContextDefinition('string')
+    ]], $this->aliasStorage);
+
+    $this->action->setStringTranslation($this->getMockStringTranslation());
+    $this->action->setTypedDataManager($this->getMockTypedDataManager());
+  }
+
+  /**
+   * Tests the summary.
+   *
+   * @covers ::summary()
+   */
+  public function testSummary() {
+    $this->assertEquals('Delete any path alias', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
+
+    $alias = 'about/team';
+
+    $this->aliasStorage->expects($this->once())
+      ->method('delete')
+      ->with(['alias' => $alias]);
+
+    $this->action
+      ->setContextValue('alias', $this->getMockTypedData($alias));
+
+    $this->action->execute();
+  }
+}


src/Plugin/Action/PathAliasDelete.php
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $language = $this->getContextValue('language');
+    $langcode = isset($language) ? $language->getId() : LanguageInterface::LANGCODE_NOT_SPECIFIED;
+
+    $source_or_alias = $this->getContextValue('source_or_alias');
+
+    $this->aliasStorage->delete($source_or_alias, $langcode);
```
klausi: I think this is wrong - AliasStorageInterface tells me that the delete() method expects a $conditions array?

src/Plugin/Action/PathAliasDelete.php
```diff
@@ -0,0 +1,107 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Delete any path alias or aliases for path' action.
+ *
+ * @Action(
+ *   id = "rules_path_alias_delete",
+ *   label = @Translation("Delete any path alias or aliases for path"),
+ *   context = {
+ *     "source_or_alias" = @ContextDefinition("string",
```
klausi: Mixing source and alias in one parameter? Why?

src/Plugin/Action/PathAliasDelete.php
stephenpurkiss: Probably - at this moment I'm learning D8 and don't fully understand the expectations of a $conditions array, examples scarce. Pointers welcome ;)

fago: According to to docs this needs a conditions passed, so I do not think this can work.

stephenpurkiss: I'm a bit lost between mock data, condition arrays, and the general mind-shift back to encapsulation so it's taking a while to figure this out - I've created a new branch (getting lost in git too;) over at https://github.com/stevepurkiss/rules/tree/pull_105_16540926_conditions_array so you can see progress and if you get chance any hints welcome ;)

src/Plugin/Action/PathAliasDelete.php
stephenpurkiss: This was when I was unsure as to the scope of this function - I've updated it to just $alias.

src/Plugin/Action/PathAliasDelete.php
stephenpurkiss: OK a good night's sleep made me realise I was getting mixed up between the string input to my class and the expected array input to the delete function on the alias storage.

tests/src/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,78 @@
...
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
+
+    // Test delete path alias
+    $alias = 'about/team';
+
+    $this->aliasStorage->expects($this->once())
+      ->method('delete')
+      ->with(array('alias' => $alias));
```
klausi: This should use short array syntax.

tests/src/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Tests the action execution when no language is specified.
+   *
+   * @covers ::execute()
+   */
+  public function testActionExecution() {
+
+    // Test delete path alias
```
klausi: Comment should end with a dot. Also the comment does not add any information, it is already clear from the function comment what we are testing?

tests/src/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,78 @@
...
+   * Tests the summary.
+   *
+   * @covers ::summary()
+   */
+  public function testSummary() {
+    $this->assertEquals('Delete any path alias', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution when no language is specified.
```
klausi: With the current implementation it is not possible to specify a language anyway?

klausi:
I think this looks mostly good and is almost ready. Could you squash your commits into one?

Then we need a follow-up to delete path aliases by source. Example: I want to delete all aliases for the path "user/1".

tests/src/Action/PathAliasDeleteTest.php
stephenpurkiss: Good point, removed.

tests/src/Action/PathAliasDeleteTest.php
stephenpurkiss: Removed text left over from previous different definition.

stephenpurkiss:
Made changes, looking into squashing now...

stephenpurkiss:
Still not getting squashing quite figured out yet but will do. In mean time what about coping with deleting an array of path aliases - or would that be a derivative?

stephenpurkiss:
Apologies for the delay, still not getting squashing. I:
1. git rebase -i HEAD~4
2. change the lines below the top to 'f'
3. it squashes the commits
4. I'm on detached head. I push using force:

durpal:rules steve$ git push origin HEAD:origin/port-delete-any-url-alias-action-to-D8-2318653 --force
Counting objects: 1, done.
Writing objects: 100% (1/1), 249 bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To https://github.com/stevepurkiss/rules.git
- f1f6fd6...00e0c90 HEAD -> origin/port-delete-any-url-alias-action-to-D8-2318653 (forced update)

...but nothing seems to change. It's probably (definitely) completely obvious what the problem is to someone who's done it before so posting here, am sure I'm just getting mixed up between branches etc.

Oh the joys of not learning this stuff earlier ;)

jpstacey:
When you rebase, imagine the last argument on the command line being the trunk end of a branch, and the "marker" for your checkout (usually a branch name) is sitting on the tip of the branch. You basically ask git to replace the whole branch down to (but not including) the trunk commit, creating a new branch with some squashed commits, and moving your "marker" over to the new tip.

So start at the tip of the commits you want to squash:

```
git checkout port-delete-any-url-alias-action-to-D8-2318653
```

And now the easiest thing to do if you're not sure is to start by creating a local "rebaser" branch - branches are cheap - and sitting on that. It means that you can always delete this local branch and start again with that first line up there.

```
git checkout -b rebaser
# or, next time around, git branch -f rebaser && git checkout rebaser
```

You then want to pick the last shared point between your commits and fago's - that is, the point you're offering the pull request onto i.e. 8.x-3.x - and rebase with that as the "trunk":

```
git rebase -i 8.x-3.x
# Make sure 8.x-3.x actually exists locally; if not, check it out first
```

Change all except one commit to "squash" or "fixup", and then when you get the option, edit the commit message so it makes sense (fixup might not give you the option).

When you're done, you should still be on `rebaser`, but `rebaser` no longer points to the same commit as it used to. So although `rebaser` and `port-delete-any-url-alias-action-to-D8-2318653` were the same, they're not any longer: `rebaser` has rebased to a stubby, single-commit branch.

So. `rebaser` is basically what fago wants at this stage. But where you go from there, with an existing pull request, I'm not sure and it might be down to fago to provide input. Normally I would check everything looks good in gitk (or gitx, or [git lol](https://gist.github.com/jpstacey/962289)) and then run:

```
# Force the local branch label over from the old branch to new rebaser
git branch -f port-delete-any-url-alias-action-to-D8-2318653
# Check that branch label out
git checkout port-delete-any-url-alias-action-to-D8-2318653
# Force-push it up to github
git push origin --force
```

However, because of this existing pull request, then force-pushing the `port-delete-any-url-alias-action-to-D8-2318653` might confuse github (this pull request could be tied to that commit at the tip of the old, discarded branch; I just don't know. I'm loath to recommend a `--force` if I don't know how remote apps are going to work.)

One option is that you could call `rebaser` something more descriptive e.g. `port-delete-any-url-alias-action-to-D8-rebased-2318653` and then offer a pull request on that, superseding this one? It depends a bit on fago's preferred workflow here, maybe?

klausi:
Github can deal just fine with rebased pull requests :)

A forced push makes the old commits disappear and only the squashed one remains.

stephenpurkiss:
OK I was rebasing to old code so I've followed the instructions but pretty sure that's wrong, but what do I know? Fairly sure the commit shouldn't have that many changes in it, just my code. Anyways, will have a play again tomorrow. Thanks again for your patience ;)

stephenpurkiss:
I think, perhaps, maybe, I've managed to sort this out on the latest commit. Understanding git a bit more now after a few more youtubes watched ;)

stephenpurkiss:
hmm.. but the build failed :(

stephenpurkiss:
Test complaining that RulesUnitTestBase not found, seem to remember RulesIntegrationTestBase being introduced so changing to that, if that passes then I get a go at rebasing again :dancers:

stephenpurkiss:
Hm maybe I rebased wrong somewhere or maybe just too far down the rabbit branch on this issue - will await someone with more knowledge than me to advise, I'm going to look into the delete aliases for path thing next.

stephenpurkiss:
...i.e. is it supposed to be unit or integration test?

stephenpurkiss:
hm ignore that one

stephenpurkiss:
ok let's try that again ;)

stephenpurkiss:
OK the travis failed on not finding unittestbase but not for my files - did I rebase wrong or is the latest commit I've done correct? (thanks and soz for the hassle, will be better next time;)

klausi:
I think you mixed up some commits here? Why do I see a ListAddItem action in the files changed?

And we still have 4 commits here, which should be squashed into one.

klausi:
The problem with the missing unit test base file is resolved now, Drupal core changed and we had to move the unit tests around.

stephenpurkiss:
Obviously messed up my local version keeping up with updates as my 8.x-3.x is reporting "This branch is 3 commits ahead of fago:8.x-3.x". Also I worked on another issue on a different branch and that's messed up in it too. Working on it ;)

stephenpurkiss:
OK, I'm back to being "This branch is even with fago:8.x-3.x"... getting there...

stephenpurkiss:
OMG I think that's it. Well, I'm going to suggest a change soon anyway from PathAliasDelete to AliasDelete because the next one I'm working on is to delete aliases for a path for which the most appropriate name is PathAliasesDelete which would cause confusion with PathAliasDelete, and as we are referring to only an alias on this issue we can drop the 'Path'. Anyway guess that's for once this is sorted. Unless you agree and I could do it in this pull request?

stephenpurkiss:
Renaming this as I am now using PathAliasDelete for deleting a path alias (https://github.com/fago/rules/pull/109), so this is now AliasDelete. Also updated path to unit tests.

stephenpurkiss:
GREEN! We like green ;)

klausi:
Finally, thanks for your endurance :-)

PR #106, opened 2014-08-23
#2281087 "text matches" condition

omissis:
Hi, I have started working on [ticket 2281087](https://www.drupal.org/node/2281087). It's far from being complete and still needs some work, but there's already something to see and I'd love to get some early feedback, so here's the PR and a few questions that I have at the moment: 
- What to do with translations? There are some untranslated strings in the exceptions I throw around the code. Also: is there a proper/common wording for exception messages?
- How about a getErrors() method in the matchers? maybe it's safer than throwing exceptions?
- I am not sure what should I pass to RulesDataMatcherPluginManager's createInstance() method yet, any suggestion?
- Is there any recommendation for functional testing the whole TextMatches class?
- DataMatcherPluginManager should be constructor-injected in TextMatches class as it's a mandatory dependency: how do I do it?
- I would like to separate the DataMatcher classes from Rules and Drupal into their own repository so to reuse it: what do you think about it?

Of course anything else that comes to your mind would be helpful, too :)


klausi:
This feels a bit heavy for Rules core just to implement the basic text comparison condition. Did you discuss this with @fago to introduce a new plugin type here?

I think we should leave this for a potential other contributed module that needs massive text comparison flexibility. A straight port of the D7 text comparison condition should suffice for now.

omissis:
I have tried to contact him on IRC but I understand he's been quite busy lately so I opted for sending in the CR and discuss here in a more asynchronous fashion.
I think a separate, plain php library (with a Drupal module wrapping it if really necessary) would be more appropriate(see last question in the comments above).
I'd prefer not to go for a direct port in that I find more useful to spend time trying to implement a cleaner, decoupled, reusable and maintainable component, so that, besides having more [SOLID](http://en.wikipedia.org/wiki/SOLID_%28object-oriented_design%29) code, not only Rules can benefit from it: in a scope wider than this issue's, we could also think of other matchers for numbers, arrays, objects and so on and so forth(I'm looking at other conditions).

klausi:
Yay, quoting software development principles! Then I would say we should follow KISS and YAGNI here. I just don't see the merit of yet another layer just to implement the most basic text comparison condition.

So I would suggest that you start a rules_matcher project (or any similar name) where you provide this implementation and where you can experiment with the matcher system and how it could replace other conditions. We can always pull it in later once we are convinced of the benefit.

omissis:
@klausi, I am sorry you interpreted my attempt at explaining what reasons drove my choices as a try to be a lecture of sort, yet irony isn't exactly helping to keep this discussion on a constructive level, nor helping me with this pull request. So let's try to keep this amicable, shall we? :)
Besides the matchers suggestions which I will indeed consider, do you have any other suggestions for the other questions? I am still not very confident with the whole system so any input would be really appreciated.

klausi:
Sorry, I might have come off as a bit snarky.

Translating exception messages: That depends on the use case. If the message will be shown to an end user on the site then translation would be in order, otherwise untranslated strings are fine. If in doubt don't translate.

throwing exceptions should be fine if you have unrecoverable errors.

creating instances: i think you need to pass the plugin ID to the manager, right?

functional testing: you could use Simpletest to write integration tests as we already do for the engine and other stuff.

injecting dependencies into plugins: you can use the static create() method pattern as for example used in CreatePathAlias.php.

separating as standalone component: I would not do it immediately since it only adds overhead during development, but rather keep everything in a rules_matcher Drupal module for now. You can still develop it as independent component in there and once it is mature move it out.

Hope that helps! :-)

fubhy:
Here is a crazy idea: Can we re-use Symfony Validator and constraints in a composable "Meets constraints" condition? I am not sure if it helps this issue but I think it's definitely worth exploring constraints for some of our simple data matching conditions.

omissis:
@fubhy that could be an idea worth looking into, I remember discussing how the symfony validators have been already used somewhere during the drupalcon amsterdam. I have also been looking at this one, which looks quite promising a probably a bit more focused on matching rather than validation: https://github.com/coduo/php-matcher.
Nevertheless, my biggest doubt here is how to trasparently include composer dependencies in a drupal module as at the time there seemed to be no clear, easy and user friendly solution. Probably now things have changed (I remember reading something about drupal 8 and composer one or two weeks ago) and if that would be the case, I'd +1 the switch to some external library.

fubhy:
We are going to work on this in a dedicated branch and then merge it into 8.x-3.x with a new PR once it's ready.

PR #107, opened 2014-08-30
Enable PHP 5.6 testing on Travis CI.

klausi:


klausi:
Drupal 8 should work now with PHP 5.6, so let's try again.

PR #109, opened 2014-09-15
Port Path Alias Delete to D8

stephenpurkiss:
Here's an initial stab at the "Delete alias for path". Unsure currently how to deal with multiple values, i.e. if there's more than one alias for the path, couldn't see a relevant function in the AliasManager for this.

### DIFF
--- a/src/Plugin/Action/PathAliasDelete.php
+++ b/src/Plugin/Action/PathAliasDelete.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\Action\PathAliasDelete.
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\Core\Path\AliasStorageInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Engine\RulesActionBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Delete alias for a path' action.
+ *
+ * @Action(
+ *   id = "rules_path_aliases_delete",
+ *   label = @Translation("Delete alias for a path"),
+ *   context = {
+ *     "path" = @ContextDefinition("string",
+ *       label = @Translation("Existing system path"),
+ *       description = @Translation("Specifies the existing path you wish to delete the alias of, for example 'node/1'. Use a relative path and do not add a trailing slash.")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add group information from Drupal 7.
+ */
+class PathAliasDelete extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The alias storage service.
+   *
+   * @var \Drupal\Core\Path\AliasStorageInterface
+   */
+  protected $aliasStorage;
+
+  /**
+   * Constructs a PathAliasDelete object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Path\AliasStorageInterface $alias_storage
+   *   The alias storage service.
+   */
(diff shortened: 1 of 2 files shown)


src/Plugin/Action/PathAliasDelete.php
```diff
@@ -0,0 +1,97 @@
...
+  public function summary() {
+    return $this->t('Delete alias for a path');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $path = $this->getContextValue('path');
+    $alias = $this->aliasManager->getAliasByPath($path);
```
klausi: Why do you need to load the alias? You can just pass the path as condition to the delete() method? That should delete all aliases for that path?

stephenpurkiss:
Why do you need to load the alias? - good question, it's my current blurred understanding of the changes from procedural to OO. It'll get there, I'm old, it takes longer to sink in these days :D Have removed references to aliasmanager and squashed the commit.

tests/src/Unit/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * @file
+ * Contains \Drupal\rules\Tests\Action\PathAliasDeleteTest.
+ */
+
+namespace Drupal\rules\Tests\Action;
+
+use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\rules\Plugin\Action\PathAliasDelete;
+use Drupal\rules\Tests\RulesUnitTestBase;
```
klausi: That file has been moved, that's why the tests are failing. Make sure to run the phunit tests locally.

stephenpurkiss:
Oops, I'd realised they'd been fixed but I thought for a moment my local 8.x-3.x branch was just out-of-date but wasn't. Fixed.

stephenpurkiss:
Right, think I've got it :D

src/Plugin/Action/PathAliasDelete.php
```diff
@@ -0,0 +1,84 @@
...
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Engine\RulesActionBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Delete alias for a path' action.
+ *
+ * @Action(
+ *   id = "rules_path_aliases_delete",
+ *   label = @Translation("Delete alias for a path"),
```
fago, line 20: Delete a path alias

src/Plugin/Action/PathAliasDelete.php
```diff
@@ -0,0 +1,84 @@
...
+use Drupal\Core\Path\AliasStorageInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Engine\RulesActionBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'Delete alias for a path' action.
+ *
+ * @Action(
+ *   id = "rules_path_aliases_delete",
```
fago, line 19: rules_path_alias_delete - it's just one

tests/src/Unit/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\rules\Plugin\Action\PathAliasDelete;
+use Drupal\Tests\rules\Unit\RulesUnitTestBase;
+
+/**
+ * @coversDefaultClass \Drupal\rules\Plugin\Action\PathAliasDelete
+ * @group rules_action
+ */
+class PathAliasDeleteTest extends RulesUnitTestBase {
```
fago, line 18: needs to be converted to the new integration test base.

tests/src/Unit/Action/PathAliasDeleteTest.php
```diff
@@ -0,0 +1,77 @@
...
+  public function testActionExecution() {
+
+    $path = 'node/1';
+
+    $this->aliasStorage->expects($this->once())
+      ->method('delete')
+      ->with(['path' => $path]);
+
+    $this->action
+      ->setContextValue('path', $this->getMockTypedData($path));
```
fago, line 73: setContextValue never receives typed data - just pass the value.

PR #110, opened 2014-09-20
Enable HHVM for testing.

klausi:


klausi:
Looks like we need to wait untill HHVM 3.3.0 is available on Travis CI.

klausi:
Drupal core does not seem ready at all for HHVM, so closing.

PR #111, opened 2014-10-01
2346659 new core adapters break tests

omissis:
PR for ticket https://www.drupal.org/node/2346659


tests/src/Unit/AutoSaveTest.php
```diff
@@ -60,13 +60,15 @@ public function testActionAutoSave() {
     $entity->expects($this->once())
       ->method('save');
 
+    $typed_data = $this->getMockTypedData($entity);
+
     $context = $this->getMock('Drupal\Core\Plugin\Context\ContextInterface');
     $context->expects($this->once())
       ->method('getContextValue')
-      ->willReturn($entity);
+      ->willReturn($typed_data);
```
fago, line 68: That's wrong, it should return the entity object.

omissis, line 68: I see but looking at Drupal\rules\Engine\RulesState:autoSave() at line 196 I see it expects a typed data object and not an entity object, probably we should solve this in there instead of here?

PR #112, opened 2014-10-01
#2317353 Conditions tests refactoring

omissis:
Ticket: 2317353 - RulesIntegrationTestBase + tests are using setContextValue() incorrectly

I have ported all the condition tests to the RulesIntegrationTest class and did some little refactorings around, such as separating Unit and Integration tests, renamed the base testcase classes to match PHPUnit's convention (*TestBase to *TestCase). I haven't looked at the Context Values yet as I am not sure if they have been fully fixed in Ticket 2346659.


tests/src/Integration/Condition/DataComparisonTest.php
```diff
@@ -2,21 +2,21 @@
...
-use Drupal\Tests\rules\Unit\RulesUnitTestBase;
+use Drupal\Tests\rules\Integration\RulesIntegrationTestCase;
+use Drupal\Tests\rules\Unit\RulesUnitTestCase;
 
 /**
  * @coversDefaultClass \Drupal\rules\Plugin\Condition\DataComparison
  * @group rules_conditions
  */
-class DataComparisonTest extends RulesIntegrationTestBase {
+class DataComparisonTest extends RulesIntegrationTestCase {
```
klausi: RulesIntegrationTestBase is named intentionally that way, because it is just an abstract base class and not an actual test case. That's how we differentiate between actual tests and helper test base classes.

tests/src/Integration/RulesIntegrationTestCase.php
```diff
@@ -134,6 +146,22 @@ public function setUp() {
   }
 
   /**
+   * Fakes the enabling of a module and loads its namespace.
+   *
+   * @param string $name the name of the module
```
fago: Description must be a whole sentence and go on the next line.

tests/src/Integration/RulesIntegrationTestCase.php
```diff
@@ -134,6 +146,22 @@ public function setUp() {
...
+   * @param array $namespaces map of the association between namespaces and filesystem paths.
```
fago: This should mention the default is fine for modules below  core/modules.

tests/src/Integration/Condition/DataComparisonTest.php
fago: Agreed.

omissis: Reverted.

klausi:
Squashed, merged and pushed. Thanks!

PR #115, opened 2014-10-03
Issue #2349413 by jetmartin: Keep using the d7 action naming pattern.

jetmartin:


src/Plugin/Action/SytemMessage.php
```diff
@@ -34,7 +34,7 @@
  * @todo: Add access callback information from Drupal 7.
  * @todo: Add group information from Drupal 7.
  */
-class DrupalMessage extends RulesActionBase {
+class SytemMessage extends RulesActionBase {
```
klausi, line 37: So we should also change the action ID in the annotation to "rules_system_message".

klausi:
Merged, thanks!

PR #120, opened 2015-01-17
Added the block ip action.

chindris:
Port "Block ip" action to D8: https://www.drupal.org/node/2317205


src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+class BlockIP extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var \Drupal\ban\BanIpManagerInterface $banManger
+   */
+  protected $banManager;
+
+  /**
+   * @param \Drupal\ban\BanIpManagerInterface $banManager
```
fago: Incomplete docs.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, BanIpManagerInterface $banManager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->banManager = $banManager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
fago, line 48: Minor, but ordering would be logical to have that on the top.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public function summary() {
+    return $this->t('Blocks an IP address');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $ip = String::checkPlain($this->getContextValue('ip'));
```
fago: Why do we need check plain to ban an ip? Seems to be not in d7 either.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,75 @@
...
+    $this->banManager->banIp($ip);
```
fago, line 86: This misses the watchdog call from d7.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,75 @@
...
+ *   context = {
+ *     "ip" = @ContextDefinition("string",
+ *       label = @Translation("IP Address")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: We should maybe use a dedicated data type for the ip address, as we
+ * do in Drupal 7.
```
fago: Also, this misses a dependency on ban module. We do not have a way to specify that yet, but let's add a todo for it.

chindris:
I fixed the issues, one thing I did not do yet, that watchdog entry. Wouldn't make more sense that the watchdog entry is created by the ban module and not by the rules module?

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,88 @@
...
+ * Provides the 'Block IP' action.
+ *
+ * @Action(
+ *   id = "rules_block_ip",
+ *   label = @Translation("Block IP"),
+ *   category = @Translation("System"),
+ *   context = {
+ *     "ip" = @ContextDefinition("string",
+ *       label = @Translation("IP Address")
+ *     )
```
klausi, line 26: In D7 the IP parameter was optional and if not provided the IP address of the current user would be used. Ithink we should do the same here again.

fago, line 26: Exactly, it should be optional - just make it "required" => false.
It should have a description that describes the fallback also.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,88 @@
...
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: We should maybe use a dedicated data type for the ip address, as we
+ * do in Drupal 7.
+ * @todo: This action depends on the ban module. We need to have a way to
+ * specify this.
+ */
+class BlockIP extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The ban manager service used to ban the IP.
```
fago, line 39: Minor, but the class does not know it's a service. So it's just The ban manager.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,88 @@
...
+   * Constructs the BlockIP object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\ban\BanIpManagerInterface $banManager
+   *   The ban manager service.
```
fago, line 67: same here.

tests/src/Integration/Action/BlockIPTest.php
```diff
@@ -0,0 +1,67 @@
...
+   */
+  protected $banManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    // We need the ban module.
+    $this->enableModule('ban');
```
fago, line 36: Do we? It looks like the class is mocked below, so it should be fine without.

tests/src/Integration/Action/BlockIPTest.php
```diff
@@ -0,0 +1,67 @@
...
+   */
+  public function testActionExecution() {
+    $ip = '192.155.122.111';
+    $this->action->setContextValue('ip', $ip);
+
+    $this->banManager->expects($this->once())
+      ->method('banIp')
+      ->with($ip);
+
+    $this->action->execute();
```
fago, line 65: Misses test coverage for the case where it takes the current IP as default.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,88 @@
...
+  public function summary() {
+    return $this->t('Blocks an IP address');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $ip = $this->getContextValue('ip');
+    $this->banManager->banIp($ip);
```
fago, line 86: Is the fallback to the current ip in ban manager?

dasjo:
There is a follow-up
https://github.com/fago/rules/pull/142

PR #122, opened 2015-01-17
Convert RulesLog to a logging service, first version. Issue #2404089

lussoluca:
Here there is a first attempt to convert the RulesLog class to a service. One thing I'm not sure is the use of Psr\Log\LogLevel levels because they're string constants and not integers, so I can't compare for level threshold.

### DIFF
--- a/src/Engine/RulesLog.php
+++ b/src/Engine/RulesLog.php
@@ -6,75 +6,89 @@
  */
 
 namespace Drupal\rules\Engine;
-use Drupal\rules\Exception;
+
+use Drupal\Core\Logger\LoggerChannelInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Routing\LinkGeneratorTrait;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\Core\Url;
+use Psr\Log\LogLevel;
 
 /**
  * The rules default logging class.
  */
-class RulesLog {
-
-  const INFO  = 1;
-  const WARN  = 2;
-  const ERROR = 3;
-
-  static protected $logger;
-
-  /**
-   * @return RulesLog
-   *   Returns the rules logger instance.
-   */
-  public static function logger($log_level = self::INFO) {
-    if (!isset(self::$logger)) {
-      $class = __CLASS__;
-      self::$logger = new $class($log_level);
-    }
-    return self::$logger;
-  }
+class RulesLog implements RulesLogInterface {
+  use LinkGeneratorTrait;
+  use StringTranslationTrait;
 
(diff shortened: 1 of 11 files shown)


tests/modules/rules_test/src/Plugin/Action/TestAction.php
```diff
@@ -24,7 +24,8 @@ class TestAction extends RulesActionBase {
    * {@inheritdoc}
    */
   public function execute() {
-    RulesLog::logger()->log('action called');
+    $logger = \Drupal::service('rules.log');
```
fubhy, line 27: This should be injected with static ::create().

templates/rules_debug_element.html.twig
```diff
@@ -0,0 +1,10 @@
+<div class="rules-debug-block">
```
fubhy, line 1: If I am reading this right this is a web profiler template. This should not be in rules but rather in the webprofiler module.

src/Engine/RulesLogInterface.php
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Gets an array of logged messages.
+   */
+  public function get();
+
+  /**
+   * Renders the whole log.
+   */
+  public function render();
```
fubhy, line 53: I think we should remove the render() part. The rules logger should simply extend AbstractLogger. Rendering should not be part of it but rather be outsourced into a separate service which can then be invoked from a specific controller which shows all log entries on a distinct page and can also be used to render the current rule log for the active request somewhere on the page.

src/Engine/RulesLog.php
```diff
@@ -6,75 +6,89 @@
...
 
 /**
  * The rules default logging class.
  */
-class RulesLog {
-
-  const INFO  = 1;
-  const WARN  = 2;
-  const ERROR = 3;
+class RulesLog implements RulesLogInterface {
```
fubhy, line 20: RulesLog should extend AbstractLogger, no?

fubhy:
I think this issue needs some fundamental conceptual planning and discussion for how we want RulesLogger to work. I think the PSR Logger (AbstractLogger) can be used directly (we need to consider a few things though). Most importantly, for allowing the same functionality we had in D7 (showing the debugger on the current page for the active request) we need to always pass a value to $context to identify the request so we can later retrieve the log entries for the current request easily. There is more to be considered though of course. I am up for discusson. You can find me on IRC or during the next Rules Hangout. Thanks for your work so far!

M1r1k:
Futher development is here #170

fubhy:
https://github.com/fago/rules/pull/170 got merged. Closing this one. Thanks for the effort (mentioned both of you in the merge commit message).

PR #123, opened 2015-01-17
First version of rules component admin ui.

dasjo:


dasjo:
https://www.drupal.org/node/2409239

rules.routing.yml
```diff
@@ -0,0 +1,11 @@
+# This file brings everything together. Very nifty!
+
+entity.rules.collection:
+  path: '/rules/list'
+  defaults:
+  # Calls the list controller, defined in the annotation of the contact entity.
```
fubhy, line 6: This is not the contact entity :).

src/Entity/Controller/RulesComponentListBuilder.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Entity\Controller\RulesComponentController.
```
fubhy, line 5: RulesComponentListBuilder

fubhy:
Thanks! Here a few things that we need to do before we can merge this:

We should add very simple version of 'edit-form' and 'delete-form' so that we at least got some operations and a form for adding rules that can then be listed.

We need functional tests. These should be rather simple and only check whether the list properly resembles a few rules that were previously added programatically or via the functional test of the add/edit form (the test can then also include the deletion). I can help with that if you need any assistance!

We need to override EntityListBuilder::load() to filter out any rules that do not have RulesComponent::$module == 'rules'. This is a (rather hidden) feature from D7 which allowed other modules to maintain and specially treat custom created rules differently and render them in a separate UI. This is something we used for Fluxkraft for example.

src/Entity/Controller/RulesComponentListBuilder.php
```diff
@@ -0,0 +1,66 @@
...
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityListBuilder;
+use Drupal\Core\Url;
+
+/**
+ * Provides a list controller for rules controller entity.
+ *
+ * @ingroup rules
+ */
+class RulesComponentListBuilder extends EntityListBuilder {
```
fubhy, line 19: Should extend ConfigEntityListBuilder instead

rules.routing.yml
```diff
@@ -0,0 +1,11 @@
+# This file brings everything together. Very nifty!
+
+entity.rules.collection:
+  path: '/rules/list'
```
fubhy, line 4: Should be /admin/config/system/rules/list instead

PR #125, opened 2015-01-18
Added the send account email action.

chindris:
Port the "Send account e-mail" action: https://www.drupal.org/node/2409061

The action execute method calls the _user_mail_notify() function. In the test class, to check if the function is called, I had to implement a special behavior of the _user_mail_notify, which I redefine in a global scope. Please check and let me know if there is a better approach.

### DIFF
--- a/tests/src/Integration/Action/SendAccountEmailTest.php
+++ b/tests/src/Integration/Action/SendAccountEmailTest.php
@@ -52,50 +52,33 @@ namespace Drupal\Tests\rules\Integration\Action {
       $this->action->setContextValue('user', $account)
         ->setContextValue('email_type', $mail_type);
 
-      // We instantiate the global mock object which will be used to call the
-      // userMailNotifyCall() method from the _user_mail_notify() in the global
-      // namespace, function which we define below.
-      global $user_mail_notify_test_helper;
-      $user_mail_notify_test_helper = $this->getMock('Drupal\Tests\rules\Integration\Action\UserMailNotifyTestHelper');
-
-      $user_mail_notify_test_helper->expects($this->once())
-        ->method('userMailNotifyCall')
-        ->with($mail_type, $account);
-
       $this->action->execute();
-    }
 
-  }
-
-  /**
-   * A helper class to check when the _user_mail_notify() is called.
-   *
-   * When testing the action, we want to check if the _user_mail_notify()
-   * global function is called. We can easy test if a method of a mock object
-   * is called, but not so easy if a php function is called. And also, the
-   * _user_mail_notify() function is not available when running the test case
-   * because the user module is not enabled. To check if the _user_mail_notify()
-   * gets called we declare it below in the global scope, and what we do is
-   * to just call the userMailNotifyCall() method of an UserMailNotifyTestHelper
-   * object (it is actually a globally available mock object).
-   *
-   * @package Drupal\Tests\rules\Integration\Action
-   */
-  class UserMailNotifyTestHelper {
-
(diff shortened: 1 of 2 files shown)


tests/src/Integration/Action/SendAccountEmailTest.php
```diff
@@ -0,0 +1,101 @@
...
+    public function testActionExecution() {
+      $account = $this->getMock('Drupal\user\UserInterface');
+      $mail_type = 'test_mail_type';
+      $this->action->setContextValue('user', $account)
+        ->setContextValue('email_type', $mail_type);
+
+      // We instantiate the global mock object which will be used to call the
+      // userMailNotifyCall() method from the _user_mail_notify() in the global
+      // namespace, function which we define below.
+      global $user_mail_notify_test_helper;
```
fubhy: This is not required. see SystemMessageTest.php

tests/src/Integration/Action/SendAccountEmailTest.php
```diff
@@ -0,0 +1,101 @@
...
+   * is called, but not so easy if a php function is called. And also, the
+   * _user_mail_notify() function is not available when running the test case
+   * because the user module is not enabled. To check if the _user_mail_notify()
+   * gets called we declare it below in the global scope, and what we do is
+   * to just call the userMailNotifyCall() method of an UserMailNotifyTestHelper
+   * object (it is actually a globally available mock object).
+   *
+   * @package Drupal\Tests\rules\Integration\Action
+   */
+  class UserMailNotifyTestHelper {
```
fubhy: Please simply add a public static method for this instead. I think that is better than having to mock this class. @fago What do you think?

EDIT: Actually you can simply use the same pattern that I used for testing drupal_set_message().

tests/src/Integration/Action/SendAccountEmailTest.php
```diff
@@ -0,0 +1,101 @@
...
+    public function userMailNotifyCall($op, $account, $langcode = NULL) {
+      // If this function gets called in this namespace, it means that the
+      // _user_mail_notify() was called.
+    }
+  }
+}
+
+namespace {
+
+  if (!function_exists('_user_mail_notify')) {
```
fubhy, line 72: This is the right approach, however as mentioned before I think it should simply call a public static method on the test class instead of using a global with a helper class that we then need to mock.

chindris:
I changed the way how to check if the _user_mail_notify() was called, and I do not use anymore the helper class. However, I did not use a static method on the test class, I do not really know how then I would check if that method was called... Is there an example somewhere? What I did now was to adapt the fake version of the _user_mail_notify() so that it stores how many times is called for each operation.

fubhy:
Hey! Yes there is an example. In fact we already have a test case for drupal_set_message() that is more or less identical to what you are trying to achieve here and you can use the same pattern. Check out SystemMessageTest.php (https://github.com/fago/rules/blob/8.x-3.x/tests/src/Integration/Action/SystemMessageTest.php).

chindris:
Yes, I saw the SystemMessageTest and actually right now the implementation is more or less similar to what is there. But I do not see now the need of the static method on the test class anymore (if you check the new implementation).

fubhy:
Sorry. I misphrased that. With static method I meant that you should put a static in that function (like you did now and like I implemented it back when we wrote the test for drupal_set_message()).

EDIT: Looks good now. Will merge tomorrow when I am at the office.

chindris:
OK, cool, thanks!

PR #129, opened 2015-01-20
#2409053 Ported data_convert action to D8 initially.

dermario:
https://www.drupal.org/node/2409053 

I am not sure if i captured all the complexity in D7 rules implementation. Any feedback is appreciated.


src/Plugin/Action/DataConvert.php
```diff
@@ -0,0 +1,79 @@
...
+   * Executes the plugin.
+   */
+  public function execute() {
+    $value = $this->getContextValue('value');
+    $target_type = $this->getContextValue('target_type');
+    $rounding_behavior = $this->getContextValue('rounding_behavior');
+
+    // First apply the rounding behavior if given.
+    if (!empty($rounding_behavior)) {
+      switch ($rounding_behavior) {
```
fubhy, line 55: I think we should throw an exception if $rounding_behavior is not one of the values supported in the switch() statement.

src/Plugin/Action/DataConvert.php
```diff
@@ -0,0 +1,79 @@
...
+        case 'round':
+          $value = round($value);
+          break;
+      }
+    }
+
+    // Avoid undefined variable notice.
+    $result = NULL;
+
+    switch ($target_type) {
```
fubhy, line 72: Same here.

src/Plugin/Action/DataConvert.php
```diff
@@ -0,0 +1,79 @@
...
+  public function execute() {
+    $value = $this->getContextValue('value');
+    $target_type = $this->getContextValue('target_type');
+    $rounding_behavior = $this->getContextValue('rounding_behavior');
+
+    // First apply the rounding behavior if given.
+    if (!empty($rounding_behavior)) {
+      switch ($rounding_behavior) {
+        case 'up':
+          $value = ceil($value);
```
fubhy, line 57: What happens here if $value is not numerical? The context definition is "any".

src/Plugin/Action/DataConvert.php
dermario, line 72: That makes sense. Should i introduce a new ContextValueException?

Edit: I just detected, that ContextException is part of Drupal Core and not part of rules. Rules itself introduces use RulesEvaluationException that might be an good way to go. Otherwise i could introduce a new RulesContextValueException or sth. like that.

fubhy, line 72: Let's summon @fago to decide that. I am not too sure about our exception strategy.

dermario:
Is there any feedback on my last pull-request or should we discuss the exception handling first?

src/Plugin/Action/DataConvert.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * @Action(
+ *   id = "rules_data_convert",
+ *   label = @Translation("Convert data"),
+ *   category = @Translation("Data"),
+ *   context = {
+ *     "value" = @ContextDefinition("any",
+ *       label = @Translation("Value")
+ *     ),
+ *     "target_type" = @ContextDefinition("any",
```
klausi: This should be "string", not "any".

klausi:
For a first port this can be just \InvalidArgumentException.

And we should also add a @todo for the action info later that was used in D7, we will have to figure out how we port hat later.

src/Plugin/Exception/RulesPluginException.php
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+namespace Drupal\rules\Plugin\Exception;
+
+use Drupal\Component\Plugin\Exception\ExceptionInterface;
+
+/**
+ * An exception class to be thrown for rules context plugin exceptions.
+ */
+class RulesPluginException extends \Exception implements ExceptionInterface { }
```
fago: Unused - should be removed.

fago:
Picked it up and completed.

PR #130, opened 2015-01-27
Action fetch entity by id 2317199

milkovsky:


src/Plugin/Action/FetchEntityById.php
```diff
@@ -0,0 +1,96 @@
...
+    return $this->t('Fetch entity by id');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $entity_type = $this->getContextValue('entity_type');
+    $entity_id = $this->getContextValue('entity_id');
+    $controller = $this->entityManager->getStorage($entity_type);
```
klausi: don't call the variable "controller", we try to avoid that to not confuse them with symfony controllers. I would suggest "storage" instead.

klausi:
Committed, thanks!

src/Plugin/Action/FetchEntityById.php
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * The entity manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+
+  /**
+   * Constructs a CreatePathAlias object.
```
milkovsky, line 51: @diarmy: Small issue in the comment for the FetchEntityById Constructor: it currently says 'Constructs a CreatePathAlias object', rather than 'Constructs a FetchEntityById object'.

tests/src/Integration/Action/FetchEntityByIdTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    $this->action
+      ->setContextValue('entity_type', 'test')
+      ->setContextValue('entity_id', 1)
+      ->execute();
+
+    // Entity load with type 'test' and id '1' should return the dummy entity.
+    $this->assertEquals($entity, $this->action->getProvided('entity')->getContextValue('entity'), 'Action returns the loaded entity for fetching entity by id.');
+  }
+}
```
milkovsky, line 93: @fago: This misses implementation and tests for providing the right metadata for the provided variable.

PR #133, opened 2015-02-05
Prepare adding a shorthand for easy Rules creation.

fago:


fago:
Nope, as discussed we put methods on the manager.

PR #134, opened 2015-02-05
Improve DX when creating and working with rule expressions

fago:
- As discussed, let's don't do shorthands.
- Attached PR caters for the 90% of adding actions and conditions by adding dedicated helpers for that, i.e. addAction() and addCondition()
- Adding general Rule expression plugins is still easy via addExpression()
- There is also addExpressionObject for the cases where you want to a full blown object. In the end that should not be necessary any more to use, but it's still useful to have and used a lot in tests. Cannot come up with a better name but I think it's ok.
- $configuration parameter is now an array of usualy expression plugin config for now. In another follow-up PR I think we should create a value object for that which improves DX and provides an API for all the expression configuration. If you think about it it's always the same (only negate for conditions differs), so we can cater at least for that common settings and then require the value object. (ContextMap, ExpressionConfig, whatever - let's discuss in detail in the next PR).
- Frankenstein-Tests are not passing yet.


klausi:
Makes all perfect sense, great work!

fago:
:)
Also created a Gist to throw out my ideas on the value object: https://gist.github.com/fago/1b6026a5fd6c203f8e19

fago:
Note: Tests are working locally, it's just the travis.yml which got broken generally.

PR #139, opened 2015-03-13
Class and interface naming improvements.

fago:
Renamed classes and interface to have consistent and meaningful naming, moved data processors below Context namespace - as they are about processing context and not necessarily part of the Engine.

Warning: This builds upon
https://github.com/fago/rules/pull/138
- thus the other one needs to be dealt wirth first.


fubhy:
What's this bullshit? There is a much more important discussion going on on Twitter atm. https://twitter.com/thefubhy/status/577476324085927936

PR #140, opened 2015-03-20
Calculate value (#2317191)

penyaskito:


src/Plugin/Action/DataCalculateValue.php
```diff
@@ -0,0 +1,91 @@
...
+ *   },
+ *  provides = {
+ *     "result" = @ContextDefinition("float",
+ *       label = @Translation("Calculated result")
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add defined operation options from Drupal 7.
```
fago, line 41: We need a todo also for better handling integers. Integers should be covered with the float action, but we need to ensure the result is an integer then also. As we cannot read the info of the actual context provided in any way yet, let's add a todo for handling integer context and results.

tests/src/Integration/Action/DataCalculateValueTest.php
```diff
@@ -0,0 +1,139 @@
...
+   * @var \Drupal\rules\Core\RulesActionInterface
+   */
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
```
fago, line 30: Unnecessary new lines here.

tests/src/Integration/Action/DataCalculateValueTest.php
```diff
@@ -0,0 +1,139 @@
...
+    $this->action->setContextValue('input_1', $this->getTypedData('float', $input_1))
+      ->setContextValue('operator', $this->getTypedData('string', 'min'))
+      ->setContextValue('input_2', $this->getTypedData('float', $input_2));
+    $this->action->execute();
+    $result = $this->action->getProvided('result')->getContextValue();
+    $this->assertEquals(min($input_1, $input_2), $result, "Min calculation correct");
+  }
+
+  /**
+   * Tests the use of php's max function for 2 input values
```
fago: Misses trailing point. Also we test the action not the php function.

tests/src/Integration/Action/DataCalculateValueTest.php
```diff
@@ -0,0 +1,139 @@
...
+    $this->action->setContextValue('input_1', $this->getTypedData('float', $input_1))
+      ->setContextValue('operator', $this->getTypedData('string', '/'))
+      ->setContextValue('input_2', $this->getTypedData('float', $input_2));
+    $this->action->execute();
+    $result = $this->action->getProvided('result')->getContextValue();
+    $this->assertEquals($input_1 / $input_2, $result, "Subtraction calculation correct");
+  }
+
+  /**
+   * Tests the use of php's min function for 2 input values
```
fago: Misses trailing point. Also we test the action not the php function.

penyaskito:
Updated with requested feedback. Didn't squash commits as I'm not sure if we want to lose track of changes, and not sure how github plays then with inline comments.

klausi:
Continued in #153

PR #141, opened 2015-03-20
Block ip 2317205

rfgpm:
Fix of the issue from the Milano Workshop


dasjo:
There is a follow-up https://github.com/fago/rules/pull/142

PR #145, opened 2015-03-21
First commit for second take on data matcher plugin.

omissis:


src/Plugin/RulesDataMatcher/ArrayContainsMatcher.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Drupal\rules\Context\DataMatcherInterface;
+
+/**
+ * Defines an array contains matcher.
+ *
+ * @RulesDataMatcher(
+ *   id = "rules_array_contains",
+ *   label = @Translation("An array contains matcher."),
+ *   subject_type = "array",
```
fago: should be list

src/Plugin/RulesDataMatcher/ArrayContainsMatcher.php
```diff
@@ -0,0 +1,31 @@
...
+ *   object_type = "array"
```
fago: is not an array, but any value. thus should be "any". generally, it should be match the data type of the list - not sure how we'd specify that - but for now it should be enough to go with list and any.

src/Plugin/RulesDataMatcher/MixedGreaterThanMatcher.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Drupal\rules\Context\DataMatcherInterface;
+
+/**
+ * Defines a 'greater than' matcher.
+ *
+ * @RulesDataMatcher(
+ *   id = "rules_mixed_greater_than",
+ *   label = @Translation("A 'greater than' matcher."),
+ *   subject_type = "mixed",
```
fago, line 18: Instead of mixed, we should use any as well. But  any would allow for objects and lists as well, so I guess we want to have "float" here? integers are usable as float also. But then, what about comparing dates? Would it make sense to support an array of types here also?

src/Plugin/RulesDataMatcher/MixedLesserThanMatcher.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Drupal\rules\Plugin\RulesDataMatcher;
+
+use Drupal\rules\Context\DataMatcherInterface;
+
+/**
+ * Defines a 'lesser than' matcher.
+ *
+ * @RulesDataMatcher(
+ *   id = "rules_mixed_lesser_than",
+ *   label = @Translation("A 'lesser than' matcher."),
```
fago, line 17: This kind of labels won't work for the select when having a form like
subject MATCHER-LABEL object
Thus, we need labels like "contains" "equals" etc. . It also shows the problem with the equality matcher - I guess it would be easier to handle if it's only one that works with every data type. Strict matching doesn't make much sense as Rules does not enforce any PHP types anyway and could be removed - so a "float" variable can be a PHP integer. -> Let's make an "any" "any" equals matcher - not sure how to add in the case sensitivity option best, but we can figure that out when we build the UI.

src/Plugin/RulesDataMatcher/MixedGreaterThanMatcher.php
omissis, line 18: I would say it would be better to pass two value objects implementing a ComparableInterface as arguments and implement it as something like this:

// throws NotComparableException if the two can't be compared
return $subject->compareTo($object);

src/Plugin/RulesDataMatcher/ArrayContainsMatcher.php
omissis: +1

src/Plugin/RulesDataMatcher/ArrayContainsMatcher.php
omissis: +1

src/Plugin/RulesDataMatcher/MixedLesserThanMatcher.php
omissis, line 17: I am a bit lost here, but I answer where I get it:

strict comparison> I suspected there might be some use cases(I'm thinking of edge cases such as string vs int or null/false/""/0 kind of values), but ok I will get rid of it;
AnyEqualsMatcher> just to make sure I understood: are you suggesting to replace StringEqualsMatcher with AnyEqualsMatcher and get rid of the former and of its case sensitive match?
labels> I am not sure I fully get the problem here, should I use the same as in rules_condition_data_is_operator_options? (ie: 'equals', 'is one of', 'is lower than', 'is greater than')

PR #146, opened 2015-03-21
Add a 'Create a new entity' action.

bojanz:
Implements a derivative plugin that provides a "Create a new $entity_type" for each content entity type.
Config entity types are not typed data aware yet, so they don't make sense in this context yet.

Disregard the summary() method, we said we want to kill it and provide it via the base class anyway.

Fixed the existing tests not to fail due to additional non-mocked method calls.


bojanz:
The tests pass locally, but since there's no test introduced for the new functionality, this should be considered broken. My initial test attempt is up at https://gist.github.com/bojanz/963b31c5a930ecde0e6a

tests/src/Integration/RulesIntegrationTestBase.php
```diff
@@ -137,6 +142,14 @@ public function setUp() {
       ->disableOriginalConstructor()
       ->getMock();
 
+    $this->entityManager = $this->getMockBuilder('Drupal\Core\Entity\EntityManagerInterface')
+      ->disableOriginalConstructor()
+      ->getMock();
+    $this->entityManager->expects($this->any())
+        ->method('getDefinitions')
```
fubhy: Indentation is wrong here :*(

tests/src/Integration/Action/FetchEntityByIdTest.php
```diff
@@ -44,6 +44,12 @@ public function setUp() {
       ])
       ->getMock();
 
+    // The base field definitions for entity_test aren't used, and would
+    // require additional mocking.
+    $this->entityManager->expects($this->any())
+        ->method('getBaseFieldDefinitions')
```
fubhy: Indentation is wrong here :*(

bojanz: Fixed, thanks. Time to quit Sublime.

src/Plugin/Action/EntityCreate.php
```diff
@@ -0,0 +1,86 @@
...
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+
+  /**
+   * The entity type.
+   *
+   * @var string
+   */
+  protected $entityType;
```
fago, line 39: Should be $entityTypeId

klausi:
continued in #158

PR #150, opened 2015-04-12
New reroll for #2189345

joshuataylor:
New reroll for https://www.drupal.org/node/2189345 .


klausi:
Ooops, was just working on this same fix and pushed it, so this should be fixed.

PR #151, opened 2015-04-13
#2317303 | Create 'Send email' action | Just rerun tests

czigor:


src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\MailManager;
+use Drupal\Core\MailManagerInterface;
+use Drupal\rules\Core\RulesActionBase;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+use Drupal\Core\Render\Element;
```
klausi: unused?

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+ * )
+ *
+ * @todo: Define that message Context should be textarea comparing with textfield Subject
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: Add group information from Drupal 7.
+ */
+class SendEmail extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var LoggerInterface $logger
```
klausi: This should use the fully qualified namespace, and the variable name should be removed. Also a description is missing. See https://www.drupal.org/coding-standards/docs#var

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+    $key = 'rules_action_mail_' . $this->getPluginId();
+
+    $message = $this->mailManager->mail('rules', $key, $send_to, $params['langcode'], $params, $reply);
+
+    if ($message['result']) {
+      $recipient = implode(", ", $send_to);
+      $this->logger->log(LogLevel::NOTICE, $this->t('Successfully sent email to %recipient', array('%recipient' => $recipient)));
+
+    }
+    return $message;
```
klausi: Why are we returning something here? The execute() method of plugins shouldn't return anything?

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+ */
+class SendEmail extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var LoggerInterface $logger
+   */
+  protected $logger;
+
+  /**
+   * @var MailManagerInterface $mailManager
```
klausi: Same here, should use proper var docs.

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+use Drupal\Core\Render\Element;
+
+/**
+ * Provides "Send email" rules action.
+ *
+ * @Action(
+ *   id = "rules_send_email",
+ *   label = @Translation("Send email"),
+ *   context = {
+ *     "send_to" = @ContextDefinition("email",
```
klausi: let's use the same context name is in D7, "to"

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+  public function summary() {
+    return $this->t('Send email');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $send_to = $this->getContextValue('send_to');
+    // @todo: Implement hook_mail_alter() in order to modify the FROM header according to https://www.drupal.org/node/2164905.
```
klausi: Not sure this todo is necessary. I think the default "reply to" approach in D8 is fine, so I Think we can remove this line.

src/Plugin/Action/SendEmail.php
```diff
@@ -0,0 +1,134 @@
...
+    $reply = $this->getContextValue('reply');
+    $params = array(
+      'subject' => $this->getContextValue('subject'),
+      'message' => $this->getContextValue('message'),
+      'langcode' => "en", //$this->getContextValue('language'),
```
klausi: This should not hardcode "en". Use the context value if available, otherwise fallback to the default language.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+
+/**
+ * @coversDefaultClass \Drupal\rules\Plugin\Action\SendEmail
+ * @group rules_actions
+ */
+class SendEmailTest extends RulesIntegrationTestBase {
+
+  protected $logger;
+  protected $mailManager;
```
klausi: var documentation missing.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+      ->setContextValue("message", 'hello');
+    $message = $this->action->execute();
+
+    $this->assertEquals(1, $message["result"]);
+    $this->assertEquals('mail@example.com', $message["to"]);
+    $this->assertEquals('admin@example.com', $message["from"]);
+
+  }
+
+}
```
klausi: newline missing.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+    $this->action->setContextValue("send_to", array('mail@example.com'))
+      ->setContextValue("subject", 'subject')
+      ->setContextValue("message", 'hello');
+    $message = $this->action->execute();
+
+    $this->assertEquals(1, $message["result"]);
+    $this->assertEquals('mail@example.com', $message["to"]);
+    $this->assertEquals('admin@example.com', $message["from"]);
+
```
klausi: You should also assert that the logger was called with the correct log message.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+  public function testSummary() {
+    $this->assertEquals('Send email', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution.
+   *
+   * @covers ::execute
+   */
+  public function testActionExecution() {
```
klausi: We should also create a test method to check that this also works with multiple receivers.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+    $params = array(
+      'to' => "mail@example.com",
+      'from' => "admin@example.com",
+      'result' => TRUE,
+    );
+
+    $this->mailManager
+      ->expects($this->any())
+      ->method('mail')
```
klausi: ->with() is missing here where you specify the expected parameters to mail() method call.

tests/src/Integration/Action/SendEmailTest.php
```diff
@@ -0,0 +1,85 @@
...
+   */
+  public function testActionExecution() {
+    $params = array(
+      'to' => "mail@example.com",
+      'from' => "admin@example.com",
+      'result' => TRUE,
+    );
+
+    $this->mailManager
+      ->expects($this->any())
```
klausi: why any()? Shouldn't this be ->once() because it should be called exactly once?

tests/src/Integration/Action/SystemSendEmailTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->logger = $this->getMock('Psr\Log\LoggerInterface');
+
+    $this->mailManager = $this->getMockBuilder('Drupal\Core\Mail\MailManagerInterface')
+      ->disableOriginalConstructor()
```
klausi: That should not be necessary since an interface does not have a constructor.

tests/src/Integration/Action/SystemSendEmailTest.php
```diff
@@ -0,0 +1,100 @@
...
+    $this->assertEquals('Send email', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution.
+   *
+   * @covers ::execute
+   */
+  public function testActionExecution() {
+    $params = array(
```
klausi: should be named $expected

src/Plugin/Action/SystemSendEmail.php
```diff
@@ -0,0 +1,135 @@
...
+    $key = 'rules_action_mail_' . $this->getPluginId();
+
+    $message = $this->mailManager->mail('rules', $key, $to, $params['langcode'], $params, $reply);
+
+    if ($message['result']) {
+      $recipient = implode(", ", $to);
+      $this->logger->log(LogLevel::NOTICE, $this->t('Successfully sent email to %recipient', array('%recipient' => $recipient)));
+
+    }
+    return $message;
```
klausi: An action does not return anything, so should be removed.

src/Plugin/Action/SystemSendEmail.php
```diff
@@ -0,0 +1,133 @@
...
+  public function summary() {
+    return $this->t('Send email');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $to = $this->getContextValue('to');
+    // @todo: Implement hook_mail_alter() in order to modify the FROM header according to https://www.drupal.org/node/2164905.
```
klausi: I think that todo can be removed. We don't have to mess around with the From header, the reply-to approach from core is fine and we should follow that.

klausi:
Merged, thanks!

PR #153, opened 2015-04-13
Calculate value 2317191

milkovsky:


src/Plugin/Action/DataCalculateValue.php
```diff
@@ -0,0 +1,92 @@
...
+  public function summary() {
+    return $this->t('Calculate a numeric value');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $input_1 = $this->getContextValue('input_1');
+    $op = $this->getContextValue('operator');
```
klausi: The variable should be called $operator.

klausi:
Merged, thanks!

PR #156, opened 2015-04-14
Block ip 2317205

ndf:
Usage of StringTranslationTrait could also be moved to RulesIntegrationTestBase . Other test do not yet make use of translation, so I expect them to only work in English.


src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,109 @@
...
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: We should maybe use a dedicated data type for the ip address, as we
+ * do in Drupal 7.
+ * @todo: This action depends on the ban module. We need to have a way to
+ * specify this.
+ */
+class BlockIP extends RulesActionBase implements ContainerFactoryPluginInterface {
```
klausi: So this is in the system category so the class and file should be prefixed with "System". Same for the test class.

tests/src/Integration/Action/BlockIPTest.php
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Tests the action execution without Context IP set.
+   *
+   * Should fallback to the current IP the user is logged in with.
+   *
+   * @covers ::execute
+   */
+  public function testActionExecutionWithoutContextIP() {
+
+    $this->action->setContextValue('ip', NULL);
```
klausi: no need to set the context with NULL, that will be set as default value anyway.

tests/src/Integration/Action/BlockIPTest.php
```diff
@@ -0,0 +1,103 @@
...
+    $ip = '12.12.12.12';
```
klausi: See https://tools.ietf.org/html/rfc3330 : "192.0.2.0/24 - This block is assigned as "TEST-NET" for use in
   documentation and example code.  It is often used in conjunction with
   domain names example.com or example.net in vendor and protocol
   documentation.  Addresses within this block should not appear on the
   public Internet."

So let's use an IP from that range.

tests/src/Integration/Action/BlockIPTest.php
```diff
@@ -0,0 +1,103 @@
...
+    $this->assertEquals('Blocks an IP address', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution with Context IP.
+   *
+   * @covers ::execute
+   */
+  public function testActionExecutionWithContextIP() {
+    $ip = '192.155.122.111';
```
klausi: See https://tools.ietf.org/html/rfc3330 : "192.0.2.0/24 - This block is assigned as "TEST-NET" for use in
   documentation and example code.  It is often used in conjunction with
   domain names example.com or example.net in vendor and protocol
   documentation.  Addresses within this block should not appear on the
   public Internet."

So let's use an IP from that range.

src/Plugin/Action/BlockIP.php
```diff
@@ -0,0 +1,109 @@
...
+ *       required = false
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ * @todo: We should maybe use a dedicated data type for the ip address, as we
+ * do in Drupal 7.
+ * @todo: This action depends on the ban module. We need to have a way to
+ * specify this.
```
davereid: You should be able to use `provider = 'ban'` in the annotation. Should also adjust the namespace to be under the ban module as well?

src/Plugin/Action/BlockIP.php
ndf: done

tests/src/Integration/Action/BlockIPTest.php
ndf: done

tests/src/Integration/Action/BlockIPTest.php
ndf: Now using:
IPv4:  '192.0.2.0'
IPv6:  '2002:0:0:0:0:0:c000:200'   <-- converted from '192.0.2.0'

tests/src/Integration/Action/BlockIPTest.php
ndf: PHPstorm is complaining that 2nd param $value is required.
Committed NULL again.

src/Plugin/Action/BlockIP.php
ndf: Talked with fago and klausi: gonna be BanIP  .
We keep the todo

src/Plugin/Action/BanIP.php
```diff
@@ -0,0 +1,109 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides the 'Ban IP' action.
+ *
+ * @Action(
+ *   id = "rules_ban_ip",
+ *   label = @Translation("Ban IP"),
+ *   category = @Translation("System"),
```
klausi: category should also be "Ban" here.

klausi:
Also the tests are failing, you forgot to rename the summary in the test I think.

src/Plugin/Action/BlockIP.php
davereid: Why keep the todo when it can be easily resolved with just `provider = "ban"`?

ndf:
Kinda silly, the build was failing on testSummary() ... 
Build is ok now.

Note that in commit "Code formatting and cleanup" I removed use statements that were not implemented. If those are related to todo's they could be added back later on.

src/Plugin/Action/BlockIP.php
ndf: Hi Dave,
Can you give an example where to put `provider = 'ban'`

Is it like this: ??

```
/**
 * Provides the 'Ban IP' action.
 *
 * @Action(
 *   id = "rules_ban_ip",
 *   label = @Translation("Ban IP"),
 *   category = @Translation("Ban"),
 *   provider = 'ban',
 *   context = {
 *     "ip" = @ContextDefinition("string",
 *       label = @Translation("IP Address"),
 *       description = @Translation("Ban an IP using the Ban Module. If no IP is provided, the current user IP is used."),
 *       required = false
 *     )
 *   }
 * )
 *

```

klausi:
merged, thanks!

PR #158, opened 2015-04-14
Entity create action

calin-marian:
Pull request for https://www.drupal.org/node/2409055


src/Plugin/Action/EntityCreate.php
```diff
@@ -0,0 +1,86 @@
...
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+
+  /**
+   * The entity type.
+   *
+   * @var string
+   */
+  protected $entityType;
```
klausi: This should be $entityTpyeId as fago said in #146

src/Plugin/Action/EntityCreate.php
```diff
@@ -0,0 +1,86 @@
...
+  public function summary() {
+    return $this->t("Create a new $this->entityType entity");
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $values = $this->getContextValues();
+    $entity = $this->entityManager->getStorage($this->entityType)->create($values);
```
klausi: So we actually don't need the entity manager but only the entity storage? Then we should inject the entity storage instead?

src/Plugin/Action/EntityCreateDeriver.php
```diff
@@ -0,0 +1,93 @@
...
+        continue;
+      }
+
+      $this->derivatives[$entity_type_id] = [
+        'label' => $this->t('Create a new @entity_type', ['@entity_type' => $entity_type->getLowercaseLabel()]),
+        'category' => $entity_type->getLabel(),
+        'entity_type' => $entity_type_id,
+        'context' => [],
+        'provides' => [
+          'entity' => new ContextDefinition('entity', $entity_type->getLabel()),
```
klausi: hm, this should be "entity:node" for example for the type?

klausi: "entity:$entity_type_id"

tests/src/Integration/Action/EntityCreateTest.php
```diff
@@ -0,0 +1,151 @@
...
+    $this->entityType = new ContentEntityType([
+      'id' => 'test',
+      'label' => 'Test',
+      'entity_keys' => [
+        'bundle' => 'bundle',
+      ],
+    ]);
+
+    // Prepare mocked entity storage.
+    $this->entityTypeStorage = $this->getMockBuilder('Drupal\Core\Entity\EntityStorageBase')
```
klausi, line 77: Any reason why we mock the concrete class instead of the interface? Usually we should mock the interface if possible.

src/Plugin/Action/EntityCreate.php
```diff
@@ -0,0 +1,86 @@
...
+ * @todo: Add access callback information from Drupal 7.
+ */
+class EntityCreate extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The entity manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
```
klausi: errr ... this should be the storage now, right?

src/Plugin/Action/EntityCreateDeriver.php
```diff
@@ -0,0 +1,93 @@
...
+        continue;
+      }
+
+      $this->derivatives["entity:$entity_type_id"] = [
+        'label' => $this->t('Create a new @entity_type', ['@entity_type' => $entity_type->getLowercaseLabel()]),
+        'category' => $entity_type->getLabel(),
+        'entity_type_id' => $entity_type_id,
+        'context' => [],
+        'provides' => [
+          'entity' => new ContextDefinition('entity', $entity_type->getLabel()),
```
klausi: the type here should be "entity:$entity_type_id"

src/Plugin/Action/EntityCreate.php
```diff
@@ -0,0 +1,86 @@
...
+      $plugin_definition,
+      $container->get('entity.manager')->getStorage($plugin_definition['entity_type_id'])
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t("Create a new $this->entityTypeId entity");
```
klausi: This should use proper placeholders with ->t(). $this->entityTypeId should go into the parameter array and should not be concatenated directly into the translatable string.

src/Plugin/Action/EntityCreateDeriver.php
```diff
@@ -0,0 +1,93 @@
...
+  public static function create(ContainerInterface $container, $base_plugin_id) {
+    return new static($container->get('entity.manager'), $container->get('string_translation'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
+    foreach ($this->entityManager->getDefinitions() as $entity_type_id => $entity_type) {
+      if (!$entity_type instanceof ContentEntityTypeInterface) {
```
klausi, line 60: This should have a commment like "Only allow content entities and ignore configuration entities."

klausi:
Merged, thanks

PR #159, opened 2015-04-14
User role add 2348839

milkovsky:


src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,71 @@
...
+    return $this->t('Adds roles to a particular user');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $account = $this->getContextValue('user');
+    $roles = $this->getContextValue('roles');
+    if ($account !== FALSE) {
```
klausi: That should not be necessary. The user context is required so we don't need to check that. The caller that invokes the action plugin is responsible to make sure that a user object is always there before executing the action.

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,71 @@
...
+          // If you try to add anonymous or authenticated role to user, Drupal
+          // will throw an \InvalidArgumentException. Anonymous or authenticated
+          // role ID must not be assigned manually.
+          $account->addRole($role->id());
+          $user_is_changed = TRUE;
+        }
+      }
+      // Save user only in case of change.
+      if ($user_is_changed) {
+        //@todo: Implement auto-save functionality. See EntitySave action.
```
klausi: Let's implement this right now.

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Tests adding of one role to user.
+   *
+   * @covers ::execute()
+   */
+  public function testAddOneRole() {
+    // Set-up a mock user.
+    $account = $this->getMockUser();
+    $account->expects($this->once())
+      ->method('addRole');
```
klausi: We should also have a ->with() call here to make sure the addRole() method is called with the correct parameter.

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Tests adding of three roles to user.
+   *
+   * @covers ::execute()
+   */
+  public function testAddThreeRoles() {
+    // Set-up a mock user.
+    $account = $this->getMockUser();
+    $account->expects($this->exactly(3))
+      ->method('addRole');
```
klausi: we should also check the parameters here to addRole()

tests/src/Integration/RulesUserIntegrationTestTrait.php
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Creates a mocked user.
+   *
+   * @return \Drupal\user\UserInterface|\PHPUnit_Framework_MockObject_MockObject
+   *   The mocked user.
+   */
+  protected function getMockUser() {
+    $account = $this
+      ->getMockBuilder('Drupal\user\Entity\User')
```
klausi: We should use the userinterface for the mock

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,69 @@
...
+        // If you try to add anonymous or authenticated role to user, Drupal
+        // will throw an \InvalidArgumentException. Anonymous or authenticated
+        // role ID must not be assigned manually.
+        $account->addRole($role->id());
+        $user_is_changed = TRUE;
+      }
+    }
+    // Save user only in case of change.
+    if ($user_is_changed) {
+      //@todo: Implement auto-save functionality. See EntitySave action.
```
milkovsky: Needs discussion.

klausi: just do it like in the EntitySave action, it is even simpler since there is no immediate saving flag.

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,69 @@
...
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Core\RulesActionBase;
+
+/**
+ * Provides a 'Add user role' action.
+ *
+ * @action(
+ *   id = "rules_user_role_add",
+ *   label = @Translation("Adds roles to a particular user"),
```
fago: Make sure to take labels and descriptions over from d7 action.

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,69 @@
...
+    $account = $this->getContextValue('user');
+    $roles = $this->getContextValue('roles');
+    // Indicates if the user account was changed.
+    $user_is_changed = FALSE;
+    foreach ($roles as $role) {
+      // Skip adding the role to the user if they already have it.
+      if (!$account->hasRole($role->id())) {
+        // For efficiency manually save the original account before applying
+        // any changes.
+        $account->original = clone $account;
```
fago: Should be done only if there is a change.

tests/src/Integration/RulesUserIntegrationTestTrait.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+trait RulesUserIntegrationTestTrait {
+
+  /**
+   * Creates a mocked user.
+   *
+   * @return \Drupal\user\UserInterface|\PHPUnit_Framework_MockObject_MockObject
+   *   The mocked user.
+   */
+  protected function getMockUser() {
```
fago: nice! should be named getMockedUser though I think.

tests/src/Integration/RulesUserIntegrationTestTrait.php
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Creates a mocked user role.
+   *
+   * @param string $name
+   *   The machine-readable name of the mocked role.
+   *
+   * @return \Drupal\user\RoleInterface|\PHPUnit_Framework_MockObject_MockBuilder
+   *   The mocked role.
+   */
+  protected function getMockUserRole($name) {
```
fago: same here.

src/Plugin/Action/UserRoleAdd.php
milkovsky: I will move it outside foreach

milkovsky: This line was here before me here. Do you know why do we need to clone it into original? I don't understand the purpose of this.

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,65 @@
...
+ * @action(
+ *   id = "rules_user_role_add",
+ *   label = @Translation("Add user role"),
+ *   category = @Translation("User"),
+ *   context = {
+ *     "user" = @ContextDefinition("entity:user",
+ *       label = @Translation("User")
+ *     ),
+ *     "roles" = @ContextDefinition("entity:user_role",
+ *       label = @Translation("Entity"),
```
klausi: "Entity"? Should be "Roles"

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,214 @@
...
+   *
+   * @covers ::execute
+   */
+  public function testAddOneRoleNoSave() {
+    // Set-up a mock user.
+    $account = $this->getMockedUser();
+    $account->expects($this->once())
+      ->method('hasRole')
+      ->with($this->equalTo('administrator'))
+      ->will($this->returnValue(FALSE));
```
klausi: addRole() on the UserInterface does not return anything, so this should be removed?

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,65 @@
...
+    $account = $this->getContextValue('user');
+    $roles = $this->getContextValue('roles');
+    foreach ($roles as $role) {
+      // Skip adding the role to the user if they already have it.
+      if (!$account->hasRole($role->id())) {
+        // If you try to add anonymous or authenticated role to user, Drupal
+        // will throw an \InvalidArgumentException. Anonymous or authenticated
+        // role ID must not be assigned manually.
+        $account->addRole($role->id());
+      }
```
klausi, line 64: So you should mark the account for auto-saving later here, meaning $this->saveLater = TRUE;

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
```
klausi, line 66: The autoSaveContext() method is missing here?

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,214 @@
...
+      ->method('save');
+
+    // Mock the 'administrator' user role.
+    $administrator = $this->getMockedUserRole('administrator');
+
+    // Test adding of one role.
+    $this->action
+      ->setContextValue('user', $account)
+      ->setContextValue('roles', [$administrator])
+      ->execute();
```
klausi, line 74: After that we need to call autoSaveContext() on the plugin and make sure that it returns ['user'].

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,214 @@
...
+      ->method('addRole');
+
+    // Mock the 'administrator' user role.
+    $administrator = $this->getMockedUserRole('administrator');
+
+    // Test adding one role.
+    $this->action
+      ->setContextValue('user', $account)
+      ->setContextValue('roles', [$administrator])
+      ->execute();
```
klausi, line 143: After that we need to call autoSaveContext() on the plugin and make sure that it does NOT return ['user'].

tests/src/Integration/Action/UserRoleAddTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+    // Mock user roles.
+    $editor = $this->getMockedUserRole('editor');
+    $administrator = $this->getMockedUserRole('administrator');
+
+    // Test adding one role.
+    $this->action
+      ->setContextValue('user', $account)
+      ->setContextValue('roles', [$administrator, $editor])
+      ->execute();
```
klausi, line 183: After that we need to call autoSaveContext() on the plugin and make sure that it returns ['user'].

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Core\RulesActionBase;
+
+/**
+ * Provides a 'Add user role' action.
+ *
+ * @action(
```
M1r1k: Shouldn't it be '@Action' ?

fago: True.

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @var bool
+   */
+  protected $saveLater = FALSE;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+    return $this->t('Adds roles to a particular user');
```
fago: outdated

src/Plugin/Action/UserRoleAdd.php
```diff
@@ -0,0 +1,78 @@
...
+    $roles = $this->getContextValue('roles');
+    foreach ($roles as $role) {
+      // Skip adding the role to the user if they already have it.
+      if (!$account->hasRole($role->id())) {
+        // If you try to add anonymous or authenticated role to user, Drupal
+        // will throw an \InvalidArgumentException. Anonymous or authenticated
+        // role ID must not be assigned manually.
+        $account->addRole($role->id());
+        // Set flag that indicates if the entity should be auto-saved later.
+        $this->saveLater = TRUE;
```
fago, line 63: Maybe init the flag at the beginning of execute() - just to be sure.

milkovsky, line 63: I think better to leave it here. If account is not changed changed we will not save it.

klausi:
Merged, thanks!

PR #160, opened 2015-04-14
Clean up naming confusion for path alias actions

mikl:
Fixes https://www.drupal.org/node/2470866

### DIFF
--- a/src/Plugin/Action/PathAliasDeleteByAlias.php
+++ b/src/Plugin/Action/PathAliasDeleteByAlias.php
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Contains \Drupal\rules\Plugin\Action\PathAliasDelete.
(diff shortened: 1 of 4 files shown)


src/Plugin/Action/PathAliasDeleteByAlias.php
```diff
@@ -16,8 +16,8 @@
  * Provides a 'Delete any path alias' action.
  *
  * @Action(
- *   id = "rules_path_alias_delete",
- *   label = @Translation("Delete any path alias"),
+ *   id = "rules_path_alias_delete_by_alias",
+ *   label = @Translation("Delete path alias"),
  *   context = {
```
klausi, line 22: category is missing here, should be "Path".

klausi:
Merged, thanks!

PR #164, opened 2015-04-15
Create a derivative for the creation of path alias for entities

nlisgo:
The d.o issue is: https://www.drupal.org/node/2471447

For the moment I have not changed \Drupal\Tests\rules\Integration\Action\NodePathAliasCreateTest much. I will continue to extend the tests for this so it is not dependant on the node entity type.

I invite feedback now on the code for \Drupal\rules\Plugin\Action\EntityPathAliasCreateDeriver and \Drupal\rules\Plugin\Action\EntityPathAliasCreate


src/Plugin/Action/EntityPathAliasCreateDeriver.php
```diff
@@ -0,0 +1,79 @@
...
+      if (!$entity_type instanceof ContentEntityTypeInterface) {
+        continue;
+      }
+
+      $this->derivatives["entity:$entity_type_id"] = [
+        'label' => $this->t('Create @entity_type path alias', ['@entity_type' => $entity_type->getLowercaseLabel()]),
+        'category' => $entity_type->getLabel(),
+        'entity_type_id' => $entity_type_id,
+        'context' => [
+          'entity' => new ContextDefinition("entity:$entity_type_id", $entity_type->getLabel(), $this->t('The entity for which to create a path alias.')),
```
fago: Let's replace "entity" here as well.

nlisgo: This has been addressed now.

nlisgo:
All of the comments from @calin-marian have now been addressed.

Please review.

tests/src/Integration/Action/EntityPathAliasCreateTest.php
```diff
@@ -0,0 +1,192 @@
...
+    $this->entityType = new ContentEntityType([
+      'id' => 'test',
+      'label' => 'Test',
+      'entity_keys' => [
+        'bundle' => 'bundle',
+      ],
+    ]);
+
+    // Prepare mocked entity manager.
+    $this->entityManager = $this->getMockBuilder('Drupal\Core\Entity\EntityManager')
```
fubhy: Why are you mocking this yourself? The RulesEntityIntegrationTestBase already does that for you.

tests/src/Integration/Action/EntityPathAliasCreateTest.php
```diff
@@ -0,0 +1,192 @@
...
+    $this->entityManager
+      ->expects($this->any())
+      ->method('getBaseFieldDefinitions')
+      ->will($this->returnValue([]));
+
+    // Return a mocked list of entity types.
+    $this->entityManager
+      ->expects($this->any())
+      ->method('getDefinitions')
+      ->will($this->returnValue(['test' => $this->entityType]));
```
fubhy: Everything you are mocking here for entity manager already exists in parent::setup().

tests/src/Integration/Action/EntityPathAliasCreateTest.php
```diff
@@ -0,0 +1,192 @@
...
+
+    // Return some dummy bundle information for now, so that the entity manager
+    // does not call out to the config entity system to get bundle information.
+    $this->entityManager
+      ->expects($this->any())
+      ->method('getBundleInfo')
+      ->with($this->anything())
+      ->will($this->returnValue(['test' => ['label' => 'Test']]));
+
+    $this->container->set('entity.manager', $this->entityManager);
```
fubhy: Everything you are mocking here for the EntityManager already exists in parent::setup().

fubhy:
Other than that this looks to be ready to go. Can you fix the setup() method (removing the entityManager mocking and the mocking of the associated methods on the entity manager?

nlisgo:
@fubhy thanks for the review. I will work on this today.

nlisgo:
@fubhy I've addressed your feedback and more. I was also able to determine that the entity_test module is enabled in RulesEntityIntegrationTestBase so we have an entity type (entity_test) that we can use removing the need to create a new one.

src/Plugin/Action/EntityPathAliasCreateDeriver.php
```diff
@@ -0,0 +1,79 @@
...
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
+    foreach ($this->entityManager->getDefinitions() as $entity_type_id => $entity_type) {
+      // Only allow content entities and ignore configuration entities.
+      if (!$entity_type instanceof ContentEntityTypeInterface) {
+        continue;
+      }
+
+      $this->derivatives["entity:$entity_type_id"] = [
```
fubhy: Not every entity type supports the path field. You need to check if the current entity type in this loop supports it and only then add it.

src/Plugin/Action/EntityPathAliasCreate.php
```diff
@@ -43,7 +32,14 @@ class NodePathAliasCreate extends RulesActionBase implements ContainerFactoryPlu
...
   /**
-   * Constructs a NodePathAliasCreate object.
+   * The entity type id.
+   *
+   * @var string
+   */
+  protected $entityTypeId;
+
+  /**
+   * Constructs an EntityCreate object.
```
klausi: Wrong class name here.

src/Plugin/Action/EntityPathAliasCreateDeriver.php
```diff
@@ -0,0 +1,79 @@
...
+    return new static($container->get('entity.manager'), $container->get('string_translation'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
+    foreach ($this->entityManager->getDefinitions() as $entity_type_id => $entity_type) {
+      // Only allow content entities with a path field.
+      if (!$entity_type instanceof ContentEntityTypeInterface || !isset($this->entityManager->getBaseFieldDefinitions($entity_type_id)['path'])) {
```
klausi: Why do we restrict this to entities with a path field? According to EntityInterface every entity has a URL, so we can also create a path alias for any entity. I think we can simply remove the second part of the condition.

tests/src/Integration/Action/EntityPathAliasCreateTest.php
```diff
@@ -0,0 +1,217 @@
...
+   */
+  protected $aliasStorage;
+
+
+  /**
+   * The mocked base field definition.
+   *
+   * @var \PHPUnit_Framework_MockObject_MockObject|\Drupal\Core\Field\BaseFieldDefinition
+   */
+  protected $baseFieldDefinition;
```
klausi: I guess this should not be necessary anymore once you don't access the base fields in the deriver anymore.

calin-marian:
fubhy said in it's previous comment that not all entity types support the path field. Therefore he suggested to check for the field.
see: https://github.com/fago/rules/pull/164#discussion-diff-28599194

klausi:
Since the path field is only a widget masking the alias storage we don't have to care at all about it :-)

calin-marian:
In this case reverting the previous commit would get us to the desired state. With the exception of some "node" keywords that were skipped.

If nlisgo can't find the time to work on this, I can take over in the weekend.

nlisgo:
@calin-marian I have taken that approach and reverted to the previous commit. Please tell me more about the "node" keywords that were skipped.

tests/src/Integration/Action/EntityPathAliasCreateTest.php
```diff
@@ -0,0 +1,144 @@
...
+   * Tests the summary.
+   *
+   * @covers ::summary
+   */
+  public function testSummary() {
+    $this->assertEquals('Create entity_test path alias', $this->action->summary());
+  }
+
+  /**
+   * Tests the action execution with an unsaved node.
```
calin-marian: I thought there were 2 instances, but it's only one. Instead of "Tests the action execution with an unsaved node" it should be "Tests the action execution with an unsaved entity".
This is in tests/src/Integration/Action/EntityPathAliasCreateTest.php at line 59

klausi:
Merged, thanks!

calin-marian:
Yay, congratulations Nathan !!!

nlisgo:
Great. I'll pick up another issue soon.

PR #167, opened 2015-04-15
Help 2470918

martin-mayer:
rules.module created and hook_help implemented. wrote documentation based on feedback from core drupal documentation lead.
https://www.drupal.org/node/2470918


dasjo:
Added feedback here:
https://www.drupal.org/node/2470918

rules.module
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * @file
+ * Allows site administrators to modify configuration.
```
klausi, line 5: That does not really fit for Rules? Should be something like "Hook implementations for the Rules module."

dasjo:
There is a newer pull request:
https://github.com/fago/rules/pull/239/

PR #169, opened 2015-04-15
Renaming RulesExpression->Expression and RulesDataProcessor->DataProcess...

stephenpurkiss:
...or


klausi:
Test fails, make sure to rename rules_expression everywhere.

stephenpurkiss:
I also changed "rulesDataProcessorManager" to "dataProcessorManager", I think that's right but would be good to double-check with you on that.

PR #173, opened 2015-04-16
Czigor/email to role action

czigor:
Tests still fail.


src/Plugin/Action/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,149 @@
...
+    $accounts = entity_load_multiple_by_properties('user', ['roles' => $rids]);
+    // @todo: Should we implement support for tokens in subject and body? in the
+    // Drupal 7 version it is not implemented for each user.
+    $params = array(
+      'subject' => $this->getContextValue('subject'),
+      'body' => $this->getContextValue('body'),
+    );
+    $from = $this->getContextValue('from');
+    if (empty($from)) {
+      $from = \Drupal::config('system.site')->get('mail');
```
M1r1k: Inject config factory

src/Plugin/Action/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,149 @@
...
+    // has been removed in the meantime).
+    $rids = array_filter(array_map(function ($role) {
+      return $role->id();
+    }, $roles));
+    if (empty($rids)) {
+      return;
+    }
+
+    // Get now all the users that match the roles (at least one of the role).
+    $accounts = entity_load_multiple_by_properties('user', ['roles' => $rids]);
```
M1r1k: Inject UserManager and use it instead of entity_load_multiple_by_properties()

src/Plugin/Action/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,149 @@
...
+      // however, $mesage['result'] is NULL, then it's likely that a module has
+      // aborted sending this particular email to this particular user, and we
+      // should just keep on sending emails to the other users.
+      if ($message['result'] === FALSE) {
+        break;
+      }
+
+    }
+    if ($message['result'] !== FALSE) {
+      $role_names = array_intersect_key(user_roles(TRUE), array_flip($roles));
```
M1r1k: Inject RoleManager

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,184 @@
...
+  public function testSummary() {
+    $this->assertEquals('Sends an e-mail to the users of a role', $this->action->summary());
+  }
+
+  /**
+   * Tests sending a mail to one role.
+   *
+   * @covers ::execute
+   */
+  public function testSendMailToOneRole() {
```
M1r1k: Use PHPUnit dataprovider instead of this huge list of actual entities and asserts

src/Plugin/Action/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,158 @@
...
+    $rids = array_filter(array_map(function ($role) {
+      return $role->id();
+    }, $roles));
+    if (empty($rids)) {
+      return;
+    }
+
+    // Get now all the users that match the roles (at least one of the role).
+
+    $accounts = entity_load_multiple_by_properties('user', ['roles' => $rids]);
```
klausi: We should inject the entity manager or storage here and query that.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Provides a 'Mail to users of a role' action.
+ *
+ * @RulesAction(
+ *   id = "rules_mail_to_users_of_role",
+ *   label = @Translation("Mail to users of a role"),
+ *   category = @Translation("System"),
+ *   context = {
+ *     "roles" = @ContextDefinition("entity",
```
klausi, line 26: This should of type "entity:user_role"

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,175 @@
...
+   * Constructs a SendMailToUsersOfRole object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Psr\Log\LoggerInterface $logger
+   *   The alias storage service.
```
klausi: alias storage?

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,175 @@
...
+   * @param \Drupal\Core\Mail\MailManagerInterface $mail_manager
+   *   The alias mail manager service.
```
klausi: should just be "The mail manager service."

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,175 @@
...
+    // Drupal 7 version it is not implemented for each user.
+    $params = array(
+      'subject' => $this->getContextValue('subject'),
+      'body' => $this->getContextValue('body'),
+    );
+    $from = $this->getContextValue('from');
+    if (empty($from)) {
+      $from = $this->config->get('mail');
+    }
+    $key = 'rules_mail_to_users_of_role_' . $this->getPluginId();
```
klausi: The plugin ID is the same as your prefix here? So I think you can just use the plugin ID alsone.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,175 @@
...
+      // should just keep on sending emails to the other users.
+      if ($message['result'] === FALSE) {
+        break;
+      }
+
+    }
+    if ($message['result'] !== FALSE) {
+      $role_names = array();
+      foreach ($roles as $role) {
+        $role_names[] = $role->id();
```
klausi: Why do you collect the role IDs here? You already have them in $rids?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+    $this->mailManager = $this->getMockBuilder('Drupal\Core\Mail\MailManagerInterface')
+      ->getMock();
+
+    // Mock the role entity.
+    $role_name = 'administrator';
+    $this->role = $this->getMockedUserRole($role_name);
+    $this->role->expects($this->any())
+      ->method('id')
+      ->willReturn($role_name);
```
klausi: This should not be necessary, the ->getMockedUserRole() method of trait already does that?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,205 @@
...
+      ->willReturn(['result' => ($call_number == 'once') ? TRUE : FALSE]);
+
+    $role_names = ($call_number == 'once') ? $this->role->id() : '';
+    $this->logger
+      ->expects($this->{$call_number}())
+      ->method('notice')
+      ->with(SafeMarkup::format('Successfully sent email to the role(s) %roles.', ['%roles' => $role_names]));
+
+    $this->action->execute();
+  }
```
klausi, line 224: So we also need a test case for multiple roles?

klausi:
ok, let's continue at the other pull request.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+    // Drupal 7 version it is not implemented.
+    $params = array(
+      'subject' => $this->getContextValue('subject'),
+      'body' => $this->getContextValue('body'),
+    );
+    $from = $this->getContextValue('from');
+    if (empty($from)) {
+      $from = $this->config->get('mail');
+    }
+    $key = $this->getPluginId();
```
klausi, line 150: Hm, the key should be something longer since the action can be used in multiple rules. Not sure how we should solve that, but please add a todo.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,238 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    $this->namespaces[] = '';
+
+    $this->logger = $this->getMock('Psr\Log\LoggerInterface');
```
klausi, line 63: Since Drupal now requires PHP 5.5 we can use LoggerInterface::class here instead of a raw string. Also elsewhere.

PR #174, opened 2015-04-16
Issue #2245031: Improve DX by invoking doExecute() and doEvaluate() methods by default.

fago:


src/Core/RulesConditionBase.php
```diff
@@ -34,4 +34,17 @@ public function negate($negate = TRUE) {
     return $this;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
```
fago, line 40: protected

PR #176, opened 2015-04-17
Data set 2409047

ndf:
See https://www.drupal.org/node/2409047

Not ready for reviewing, but added a DataSetTest and a summary function.


klausi:
Continued in #202

PR #178, opened 2015-04-17
page redirect action 2471641

bbujisic:
See https://www.drupal.org/node/2471641

There is a pretty large list of todo's:
- We need a "restriction" context.
- ~~We need an "optional" context~~ We have required=false now.
- We need a "defaultValue" context.
- There should be a line of code preventing redirection rule from being executed on rules admin pages (in order not to lock out admins). But we still do not have the URL to rules admin page...
- Lines 124-128 of src/Plugin/Action/SystemPageRedirect.php are absolutely untested and I have no idea if they are going to work.
- Lines 131-133 of src/Plugin/Action/SystemPageRedirect.php are also untested and I have no idea if they are going to work.
- More test scenarios might be needed.


src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,138 @@
...
+    if (isset($batch['current_set'])) {
+      $this->logger->log(LogLevel::WARNING, $this->t('Skipped page redirect during batch processing'));
+      return;
+    }
+
+    // Keep the current destination parameter if there is one set.
+    // @todo: This might not work. Test!
+    if ($keepDestination) {
+      $url .= strpos($url, '?') === FALSE ? '?' : '&';
+      $url .= \Drupal::destination()->get();
```
fubhy: Please inject the 'redirect.destination' service for this.

tests/src/Integration/Action/SystemPageRedirectTest.php
```diff
@@ -0,0 +1,85 @@
...
+   */
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+
+    require_once __DIR__ . '/../../../../../../core/includes/form.inc';
```
fubhy: Aha. I understand your problem now. Find me in the rules sprint room so I can show you how to do this.

src/Plugin/Action/SystemPageRedirect.php
bbujisic: This is a redirect.destination service now.

tests/src/Integration/Action/SystemPageRedirectTest.php
bbujisic: As advised, this is fixed by mocking batch_set() and batch_get() in test file.

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+ *   }
+ * )
+ *
+ * @todo: Check if we can use context.url.type = "uri".
+ * @todo: Check if we can use context.force.restriction = "input".
+ * @todo: Check of we can use context.force.optional = true.
+ * @todo: Check of we can use context.force.defaultValue = true.
+ * @todo: Check if we can use context.destination.restriction = "input".
+ * @todo: Check of we can use context.destination.optional = true.
+ * @todo: Check of we can use context.destination.defaultValue = true.
```
milkovsky: typo. "if"

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+    $force = $this->getContextValue('force');
+    $keepDestination = $this->getContextValue('destination');
+
+    // @todo: We need URL to Rules administration pages in order to make sure administrators won't get locked out from them.
+
+    // Make sure we do not redirect during batch processing.
+    $batch = batch_get();
+    if (isset($batch['current_set'])) {
+      $this->logger->log(LogLevel::WARNING, $this->t('Skipped page redirect during batch processing'));
+      $this->setProvidedValue('redirect', false);
```
milkovsky: false, true, null should be upper case. FALSE, TRUE, NULL. Please check everywhere

tests/src/Integration/Action/SystemPageRedirectTest.php
```diff
@@ -0,0 +1,121 @@
...
+      /* @var \Symfony\Component\HttpFoundation\RedirectResponse $redirect */
+      $redirect = $this->action->getProvidedContext('redirect')
+        ->getContextValue();
+
+      $this->assertEquals($redirect, false);
+    }
+  }
+}
+
+namespace {
```
milkovsky, line 113: wow. what is this? haven't seen such code so far

bbujisic, line 113: This was actually fubhy's suggestion at DevDays Montpellier: batch_get function is a part of /core/includes/form.inc and cannot be reliably loaded from within module's directory (include would look like this: include(__ DIR __. '/../../../../../core/.../form.inc')..). Is there a better way to achieve this?

klausi, line 113: This is fine as is, we can only test functions reliably like this in PHPUnit.

fubhy, line 113: @milkovsky Instead of declaring the namespace globally for the file it wraps only the class itself in it, allowing us to define global namespace functions (batch_set/batch_get) below which we can then use in the unit test without loading the actual implementations (which would destroy the unit test). Basically, this causes the class tested by this unit test to use our implementation of batch_set/batch_get when the respective code is reached. It's a neat little trick we are using elsewhere in Rules already. I think I came up with that when writing the unit test for drupal_set_message, IIRC.

fubhy, line 113: Yeah, that was here: https://github.com/fago/rules/pull/75/files

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+ *   provides = {
+ *     "redirect" = @ContextDefinition("any",
+ *       label = @Translation("Redirect")
+ *     ),
+ *   }
+ * )
+ *
+ * @todo: Check if we can use context.url.type = "uri".
+ * @todo: Check if we can use context.force.restriction = "input".
+ * @todo: Check if we can use context.force.optional = TRUE.
```
klausi: That is already possible, you just need to set required = false for that context. See the EntitySave action for an example.

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Provides "Page redirect" rules action.
+ *
+ * @Action(
+ *   id = "rules_page_redirect",
+ *   label = @Translation("Page redirect"),
+ *   category = @Translation("System"),
+ *   context = {
+ *     "url" = @ContextDefinition("uri",
```
klausi: I think this should be "string" since a Drupal path is not a valid uri, but can still be used here.

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+ *     ),
+ *   }
+ * )
+ *
+ * @todo: Check if we can use context.url.type = "uri".
+ * @todo: Check if we can use context.force.restriction = "input".
+ * @todo: Check if we can use context.force.optional = TRUE.
+ * @todo: Check if we can use context.force.defaultValue = TRUE.
+ * @todo: Check if we can use context.destination.restriction = "input".
+ * @todo: Check if we can use context.destination.optional = TRUE.
```
klausi: Same here, just set required = false.

src/Plugin/Action/SystemPageRedirect.php
```diff
@@ -0,0 +1,141 @@
...
+ *     "force" = @ContextDefinition("boolean",
+ *       label = @Translation("Force redirect"),
+ *       description = @Translation("Force the redirect even if another destination parameter is present. Per default Drupal would redirect to the path given as destination parameter, in case it is set. Usually the destination parameter is set by appending it to the URL."),
+ *     ),
+ *     "destination" = @ContextDefinition("boolean",
+ *       label = @Translation("Append destination parameter"),
+ *       description = @Translation("Whether to append a destination parameter to the URL, so another redirect issued later on would lead back to the origin page."),
+ *     ),
+ *   },
+ *   provides = {
```
klausi: I don't think this action should provide something. It should only order the redirect.

klausi: So as far as I can see we will have to create a Response listener that will issue the redirect if it was ordered by the action execution.

bbujisic: Okay, I googled around and bothered a couple of Finns in DrupalcampCS in order to figure out how to do this Response listener thing. The nearest thing to what we need was http://cgit.drupalcode.org/globalredirect/tree/src/EventSubscriber/GlobalredirectSubscriber.php?h=8.x-1.x.

I got stuck with one problem and one question.

The problem: if we create event subscriber class which implements EventSubscriberInterface and handles redirections and add it through rules.services.yml, how will it know when was the rule action executed?

The question is simpler -- is this an overkill? Or is this redirection thing responsibility of core?

bbujisic:
Update: I cloned the way core Drupal\action\Plugin\Action\GotoAction adds redirection listeners.

Tests are a little bit ugly because they only assert if response listener exists in event dispatcher. I have no idea how to access the RedirectResponse object within the listener to assert if response is ok.

Help please?

klausi:
Committed with #295

PR #179, opened 2015-04-17
Fixed travis.yml as #2467411 was commited to the core

lu4nik:
https://www.drupal.org/node/2467411#comment-9836183 - committed to the core, travis test are falling now because of this.


fubhy:
Thanks.

PR #182, opened 2015-04-17
port_remove_user_role-2471645-6.patch

katzilla:
Adding a patch for https://www.drupal.org/node/2471645
Added RulesUserIntegrationTestTrait again - might be doubled now, because a.milkovsky uploaded before, but it wasn't merged yet...
Please review - it's my first unittest.... ;-)

### DIFF
--- a/src/Plugin/Action/UserRoleRemove.php
+++ b/src/Plugin/Action/UserRoleRemove.php
@@ -2,6 +2,7 @@
 
 
 namespace Drupal\rules\Plugin\Action;
+
 use Drupal\rules\Core\RulesActionBase;
 
 /**
@@ -56,9 +57,7 @@ class UserRoleRemove extends RulesActionBase
     foreach ($roles as $role) {
       // Check if user has role.
       if ($account->hasRole($role->id())) {
-        // If you try to add anonymous or authenticated role to user, Drupal
-        // will throw an \InvalidArgumentException. Anonymous or authenticated
-        // role ID must not be assigned manually.
+        // Remove role
         $account->removeRole($role->id());
         // Set flag that indicates if the entity should be auto-saved later.
         $this->saveLater = TRUE;
--- a/src/Plugin/Action/UserRoleRemove.php
+++ b/src/Plugin/Action/UserRoleRemove.php
@@ -1,6 +1,5 @@
 <?php
 
-
 namespace Drupal\rules\Plugin\Action;
 
 use Drupal\rules\Core\RulesActionBase;
@@ -41,7 +40,6 @@ class UserRoleRemove extends RulesActionBase
   {
     return $this->t('Remove roles of particular users');
   }
-
 
   /**
    * {@inheritdoc}
--- a/src/Plugin/Action/UserRoleRemove.php
+++ b/src/Plugin/Action/UserRoleRemove.php
@@ -1,0 +1,79 @@
+<?php
+
+
+namespace Drupal\rules\Plugin\Action;
+use Drupal\rules\Core\RulesActionBase;
+
+/**
+ * Provides a 'Remove user role' action.
+ *
+ * @action(
+ *   id = "rules_user_role_remove",
+ *   label = @Translation("Remove user role"),
+ *   category = @Translation("User"),
+ *   context = {
+ *     "user" = @ContextDefinition("entity:user",
+ *       label = @Translation("User")
+ *     ),
+ *     "roles" = @ContextDefinition("entity:user_role",
+ *       label = @Translation("Roles"),
+ *       multiple = TRUE
+ *     )
+ *   }
+ * )
+ *
+ */
+class UserRoleRemove extends RulesActionBase
+{
+
+  /**
+   * Flag that indicates if the entity should be auto-saved later.
+   *
+   * @var bool
+   */
+  protected $saveLater = FALSE;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary()
+  {
+    return $this->t('Remove roles of particular users');
+  }
+
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute()
+  {
+
+    /** @var \Drupal\user\Entity\User $account */
+    $account = $this->getContextValue('user');
+
+    /** @var \Drupal\user\RoleInterface $roles */
+    $roles = $this->getContextValue('roles');
+    foreach ($roles as $role) {
+      // Check if user has role.
+      if ($account->hasRole($role->id())) {
+        // If you try to add anonymous or authenticated role to user, Drupal
+        // will throw an \InvalidArgumentException. Anonymous or authenticated
+        // role ID must not be assigned manually.
+        $account->removeRole($role->id());
+        // Set flag that indicates if the entity should be auto-saved later.
+        $this->saveLater = TRUE;
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function autoSaveContext()
+  {
+    if ($this->saveLater) {
+      return ['user'];
+    }
+    return [];
+  }
+}
--- a/src/Plugin/Action/UserRoleRemove.php
+++ b/src/Plugin/Action/UserRoleRemove.php
@@ -1,13 +1,18 @@
 <?php
+/**
+ * @file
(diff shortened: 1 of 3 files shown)


src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace Drupal\rules\Plugin\Action;
+
+use Drupal\rules\Core\RulesActionBase;
+
+/**
+ * Provides a 'Remove user role' action.
+ *
+ * @action(
```
milkovsky: Should be @Action

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+ *     "roles" = @ContextDefinition("entity:user_role",
+ *       label = @Translation("Roles"),
+ *       multiple = TRUE
+ *     )
+ *   }
+ * )
+ *
+ */
+class UserRoleRemove extends RulesActionBase
+{
```
fubhy: Should be on the same line as the class name (@see Drupal Coding Standards).

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @var bool
+   */
+  protected $saveLater = FALSE;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary()
+  {
```
fubhy: Opening Curly Brackets should always be on the same line as the name of the class / method / function

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function execute()
+  {
+
+    /** @var \Drupal\user\Entity\User $account */
+    $account = $this->getContextValue('user');
+
+    /** @var \Drupal\user\RoleInterface $roles */
```
fubhy: This is a LIST of roles so if you want to use the @var notation then you will have to specify it as 

```
\Drupal\user\RoleInterface[]
```

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+    return $this->t('Remove roles of particular users');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute()
+  {
+
+    /** @var \Drupal\user\Entity\User $account */
```
fubhy: Pleas take a look at the RulesActionBase implementation. Wolfgang commited a DX improvement patch that now lets you use a doExecute() method instead where the arguments are the defined contexts (in the order in which they are defined) so you can use direct type hinting there.

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+    $account = $this->getContextValue('user');
+
+    /** @var \Drupal\user\RoleInterface $roles */
+    $roles = $this->getContextValue('roles');
+    foreach ($roles as $role) {
+      // Check if user has role.
+      if ($account->hasRole($role->id())) {
+        // Remove role
```
fubhy: These comments are redundant given the expressive method name and API usage.

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+        $account->removeRole($role->id());
+        // Set flag that indicates if the entity should be auto-saved later.
```
fubhy: Same here. Pretty self explanatory from looking at the code. We should try to avoid redundant comments (which should be true for all of Drupal, we tend to overdo these things).

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
...
+   * {@inheritdoc}
+   */
+  public function autoSaveContext()
+  {
+    if ($this->saveLater) {
+      return ['user'];
+    }
+    return [];
+  }
+}
```
fubhy: Missing line break at end of file.

tests/src/Integration/Action/UserRoleRemoveTest.php
```diff
@@ -0,0 +1,110 @@
...
+      )
+      ->will($this->returnValue(FALSE));
+
+    $account2->expects($this->never())
+      ->method('removeRole')
+      ->with(
+        $this->equalTo('editor')
+      );
+
+    return array(
```
fubhy, line 101: We are using the short array syntax throughout. Please switch to [] instead of array().

tests/src/Integration/Action/UserRoleRemoveTest.php
```diff
@@ -0,0 +1,110 @@
...
+      'removing of one role' => array($account1, ['user']),
```
fubhy: Same here and in the line below.

tests/src/Integration/RulesUserIntegrationTestTrait.php
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @file
+ * Contains \Drupal\Tests\rules\Integration\RulesUserIntegrationTestTrait.
+ */
+
+namespace Drupal\Tests\rules\Integration;
+
+/**
+ * Base class for Rules integration tests with user entities.
```
fubhy: This is not a base class but a trait :).

tests/src/Integration/RulesUserIntegrationTestTrait.php
```diff
@@ -0,0 +1,45 @@
...
+      ->getMock();
+
+    $role->expects($this->any())
+      ->method('id')
+      ->will($this->returnValue($name));
+
+    return $role;
+  }
+
+}
```
fubhy: Missing new line at end of file.

tests/src/Integration/Action/UserRoleRemoveTest.php
```diff
@@ -0,0 +1,110 @@
...
+  {
+    $this->assertEquals('Remove roles of particular users', $this->action->summary());
+  }
+
+  /**
+   * Tests removing existing role from user.
+   *
+   * @covers ::execute
+   * @dataProvider rolesProvider
+   * @param $account
```
fubhy: While it annoys me that we have to do this, it violates the Drupal Coding Standards to not document the parameters of a method/function with a short description.

tests/src/Integration/Action/UserRoleRemoveTest.php
```diff
@@ -0,0 +1,110 @@
...
+   * @param $user
```
fubhy: Same here.

tests/src/Integration/Action/UserRoleRemoveTest.php
```diff
@@ -0,0 +1,110 @@
...
+    $this->action
+      ->setContextValue('user', $account)
+      ->setContextValue('roles', [$editor])
+      ->execute();
+
+    $this->assertEquals($this->action->autoSaveContext(), $user, 'Action returns nothing for auto saving since the user has been saved already.');
+  }
+
+  /**
+   * @return array
```
fubhy, line 70: Please add a short description "Provides mocked user objects for the testRemoveExistingRole test." or somtehing like that.

fubhy:
Thanks, this is really looking good. Just some nitpicks as you can see from my comments. Would be great if you could fix those :). Thanks!

src/Plugin/Action/UserRoleRemove.php
pjezek: ok, we will to fix that in 2470661 and 2470665 as well.

pjezek: was that announced in the irc channel?

src/Plugin/Action/UserRoleRemove.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
```
dasjo, line 6: There should be a @file comment of the beginning of each file. Please check other Action plugins for reference.

stephenpurkiss:
I have continued work on this over at PR 214 https://github.com/fago/rules/pull/214

klausi:
OK, let's continue there.

PR #183, opened 2015-04-19
Add a Lexer/Parser for mathematical expressions. (Input Processor + Separate Action & Condition)

fubhy:
Fixing implementation of the Shunting-yard algorithm.


fubhy:
This is not yet finished. It is lacking test coverage and the usage of plain arrays for the output and for the stack in the Shunting-yard algorithm is going to be subject to change (I want to use \SqlQueue and \SqlStack there... Same for the initial tokenizing).

Note: This implementation seems to be extremely effective due to the compiled regex. I tried running a few (even complex) mathematical expressions and they correctly evaluated within a matter of microseconds. Yai

fubhy:
Oh and it obviously lacks Drupal / Rules integration (still misses variable support, currently only handles constants). I also want to register the lexer and the parser as separate services. Not sure yet how I want to handle the registration of the default methods. Probably through a list of method calls in the services.yml.

src/Math/Calculator.php
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Drupal\rules\Math;
+
+use Drupal\rules\Math\Exception\IncorrectExpressionException;
+use Drupal\rules\Math\Token\ConstantToken;
+use Drupal\rules\Math\Token\FunctionToken;
+use Drupal\rules\Math\Token\NumberToken;
+use Drupal\rules\Math\Token\OperatorTokenInterface;
+
+class Calculator {
```
dasjo, line 20: Add a docblock

src/Math/Calculator.php
```diff
@@ -0,0 +1,96 @@
...
+      ->addFunction('sinh', 'sinh', 1)
+      ->addFunction('sqrt', 'sqrt', 1)
+      ->addFunction('tan', 'tan', 1)
+      ->addFunction('tanh', 'tanh', 1);
+
+    $this->lexer->addConstant('pi', pi());
+    $this->lexer->addConstant('e', exp(1));
+  }
+
+  public function calculate($expression, $variables) {
```
dasjo, line 92: Let's document this

src/Math/Lexer.php
```diff
@@ -0,0 +1,319 @@
...
+
+    if (preg_match_all($regex, $input, $matches, PREG_SET_ORDER | PREG_OFFSET_CAPTURE) === FALSE) {
+      throw new \LogicException();
+    };
+
+    $types = [
+      'number', 'operator', 'function', 'open', 'close', 'comma', 'constant',
+    ];
+
+    return array_map(function ($match) use ($types) {
```
dasjo, line 136: Looks like beautiful magic to me. I think it would be great to describe whats the data structure should look like. As we don't have interfaces describing stuff, maybe the methods should be a bit more verbose on what they are doing

src/Math/Token/TokenInterface.php
```diff
@@ -0,0 +1,15 @@
...
+ * Contains \Drupal\rules\Math\Token.
+ */
+
+namespace Drupal\rules\Math\Token;
+
+/**
+ * Provides a common interface for all tokens.
+ */
+interface TokenInterface {
+
```
dasjo, line 14: Add the methods from the token trait here?

src/Math/Calculator.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace Drupal\rules\Math;
```
dasjo, line 8: Needs @file

src/Math/Calculator.php
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public function calculate($expression, $variables) {
+    $hash = md5($expression);
+    if (isset($this->tokenCache[$hash])) {
+      return $this->tokenCache[$hash];
+    }
+
+    $stream = $this->lexer->tokenize($expression);
+    $tokens = $this->lexer->postfix($stream);
```
dasjo, line 99: So after the postfix, tokens will be in reverse polish notation. Which is the reason why you can use a stack to evaluate. Would be cool to explain this a bit more

src/Math/Lexer.php
```diff
@@ -0,0 +1,319 @@
+<?php
+
+namespace Drupal\rules\Math;
```
dasjo, line 8: Needs @file

src/Math/Lexer.php
```diff
@@ -0,0 +1,319 @@
...
+          $output[] = array_pop($stack);
+        }
+
+        array_push($stack, $token);
+      }
+    }
+
+    while (!empty($stack)) {
+      $token = array_pop($stack);
+      if ($token instanceof ParenthesesOpenToken || $token instanceof ParenthesesCloseToken) {
```
dasjo: Could you add a comment why at this stage no parantheses are valid?

src/Math/Token/TokenInterface.php
fubhy, line 14: I removed the token trait now.

src/Math/Lexer.php
```diff
@@ -0,0 +1,324 @@
...
+   * @param string $name
+   *   The name of the function.
+   * @param callable $function
+   *   The callback to be executed.
+   * @param int $arguments
+   *   The number of arguments required for the function.
+   *
+   * @return $this
+   */
+  public function addFunction($name, callable $function, $arguments = 1) {
```
dawehner, line 69: Can't you figure out the amount of arguments using reflection?

src/Math/Lexer.php
```diff
@@ -0,0 +1,324 @@
...
+   *   The name of the operator.
+   * @param string $regex
+   *   The regular expression of the operator token.
+   * @param string $operator
+   *   The full qualified class name of the operator token.
+   *
+   * @return $this
+   */
+  public function addOperator($name, $regex, $operator) {
+    if (!in_array('Drupal\rules\Math\Token\OperatorTokenInterface', class_implements($operator))) {
```
dawehner: you probably can use is_subclass_of

src/Math/Lexer.php
fubhy, line 69: I would like to avoid reflection due to performance concerns. Do you think it is viable to do this on runtime? Or are you saying that, once I move this to some other place, I would do this during the building of the container?

dawehner, line 69: Fair point. Yeah I kinda assumed that you cache that object

fubhy, line 69: Well, I will... Eventually (once I am registering as a service). So I will look at that again once that's done. That huge constructor with all the registering is going to go away in that process of course (that's just for now so I can work with it and test it).

fago, line 69: Makes sense. The variable should be number_of_arguments or so though, as it's not the actual function arguments.

fubhy, line 69: Once we retrieve that with reflection that argument would be gone anyways :P.

fubhy:
I just made this a separate PHP library on GitHub: https://github.com/fubhy/math-php

PR #184, opened 2015-04-23
Added code for refining item context

stephenpurkiss:
Initial code for refining item context. See https://www.drupal.org/node/2281083#comment-9834577

Posting for visibility and testing.


src/Plugin/Condition/DataListContains.php
```diff
@@ -39,6 +39,20 @@ class DataListContains extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
+  public function refineContextdefinitions() {
+
+    // Refine the item context
+    if($item = $this->getContextValue('item')) {
```
klausi: there should be one space after the "if"

tests/src/Integration/Condition/ListContainsTest.php
```diff
@@ -139,4 +139,30 @@ public function testConditionEvaluation() {
...
+
+  /**
+   * @covers ::refineContextDefinitions
+   */
+  public function testRefiningContextDefinitions() {
+    // Create array of mock entities
+    $entity_zero = $this->getMock('Drupal\Core\Entity\EntityInterface');
+    $entity_zero->expects($this->any())
+      ->method('id')
+      ->will($this->returnValue('entity_zero_id'));
```
klausi: can be simplified to just "->willReturn()"

tests/src/Integration/Condition/ListContainsTest.php
```diff
@@ -139,4 +139,30 @@ public function testConditionEvaluation() {
...
+    
+    $entity_one = $this->getMock('Drupal\Core\Entity\EntityInterface');
+    $entity_one->expects($this->any())
+      ->method('id')
+      ->will($this->returnValue('entity_one_id'));
```
klausi: can be simplified to just "->willReturn()"

src/Plugin/Condition/DataListContains.php
```diff
@@ -39,6 +39,20 @@ class DataListContains extends RulesConditionBase {
...
+      if ($item instanceof EntityInterface && $type = $item->getTypedData('type')) {
```
klausi: ->getTypeData() will return an object, but we want the entity type ID here, so it should be getEntityTypeId().

tests/src/Integration/Condition/ListContainsTest.php
```diff
@@ -139,4 +139,30 @@ public function testConditionEvaluation() {
...
+
+    // Test array of entities
+    $entity_list = [$entity_zero, $entity_one];
+
+    $this->condition
+      ->setContextValue('list', $entity_list)
+      ->setContextValue('item', $entity_zero);
+    $this->condition->refineContextdefinitions();
+    $this->assertEquals(
+      $this->condition->getContextValue('item'), $entity_zero);
```
klausi: Actually we should get the context definition here for the item and make sure that the type is now the entity type.

stephenpurkiss:
Will try to get on the regular IRC meeting, meanwhile here's where I'm at on this:

https://www.drupal.org/node/2281083#comment-9866613

So here's where my thinking's at currently:
- from what I can tell, 7 allows you to alter the type of a list item. It uses the same alter hook for many similar functions such as add item to list, list contains, etc. With my limited knowledge of form api I can't tell whether it allows for the type of the item being added to the list (or whatever the function is) to be altered or not, but my hunch is it doesn't. Should it? That's another question - if alter hooks are there for enabling the complete control over types at runtime then yes, it should.
- in 8 we should allow for refining configs on all available places, i.e. type of item and type of each item in list. I don't currently know how to achieve this in code, hence why posting update here before I spend days figuring out something which is wrong ;) It does beg the question though that surely if all this alter 'hook' is allowing for is changing of type then it could be extracted to a level higher up as 'type' should be discoverable?

Still reading up on various discussions around this so no doubt more will become clear as has done in the past week!

stephenpurkiss:
OK I think I've got this working thanks to @tstoeckler who was here in spain today. 

My next question is this covers refining the context for the item, not the list. In 7 you couldn't alter the item context, only the list. So I'm presuming I need to add code to alter context of the list? Do I need to step through each item in the list to achieve this? Some pointer would be good - I'm done for the day here but will continue tomorrow/next week.

src/Plugin/Condition/DataListContains.php
```diff
@@ -39,6 +39,20 @@ class DataListContains extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
+  public function refineContextDefinitions() {
+
+    // Refine the item context.
+    if ($item = $this->getContextValue('item')) {
+      if ($item instanceof EntityInterface && $type = $item->getEntityTypeId()) {
```
klausi: why EntityInterface here? I think this should just be "$typed_data_item = $this->getContextData('item')" and then "$this->pluginDefinition['context']['item']->setDataType($typed_data_item->getDataDefinition()->getDataType());"

stephenpurkiss: "$typed_data_item = $this->getContextData('item')" created this error:

PHP Fatal error:  Call to undefined method Drupal\rules\Plugin\Condition\DataListContains::getContextData() in /Users/steve/Sites/d8dev/modules/rules/src/Plugin/Condition/DataListContains.php on line 45

...so I copied the code from DataComparison.php:

$item = $this->condition->getContextDefinition('item');

Your suggested code failed my test though:

Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'any'
+'entity:entity_zero_type_id'

tests/src/Integration/Condition/ListContainsTest.php
```diff
@@ -139,4 +139,42 @@ public function testConditionEvaluation() {
...
+    $entity_list = [$entity_zero, $entity_one];
+
+    $this->condition
+      ->setContextValue('list', $entity_list)
+      ->setContextValue('item', $entity_zero);
+    $this->condition->refineContextdefinitions();
+
+    // Get the context definition for the item
+    // and make sure that the type is now the entity type.
+    $item = $this->condition->getContextDefinition('item');
```
klausi: item is a bad variable name here, since this is not the item but the context definition of item.

stephenpurkiss: Good point, well made ;)

stephenpurkiss:
OK I've done the changes recommended but now stuck at what needs mocking in the test. I realise now we're not using getDataTypeId but getDataType so changed that but still no joy. Tried mocking all the things - getContext, getContextData and getDataDefinition, but that made no difference, still coming back as 'any' type.

stephenpurkiss:
Am looking at ContextTest.php and see a whole setup routine for mocking context definitions and can see it's a bit more involved so will investigate that further

PR #188, opened 2015-05-05
Config export

LKS90:
Exporting configurations is now supported and helps with performance when creating an array from the schema file. [Since a recent change](https://www.drupal.org/node/2481909), ConfigEntityType annotation can declare an array of config entities to export, which saves a lot of time, for example during site install.


klausi:
While config entity support in Rules is still very much unfinished this does not hurt and we will take it into account for future changes.

Merged, thanks!

PR #193, opened 2015-05-31
Allow execution expression containers with any defined context

klausi:
See https://www.drupal.org/node/2453545


klausi:
Merged.

PR #194, opened 2015-06-05
Implement Rules Events

klausi:
See https://www.drupal.org/node/2145343


klausi:
merged

config/schema/rules.schema.yml
```diff
@@ -29,6 +29,38 @@ rules.component.*:
...
+      label: 'Module'
+    description:
+      type: text
+      label: 'Administrative description'
+    tag:
+      type: string
+      label: 'Tag'
+    core:
+      type: string
+      label: 'Drupal version'
```
loopdrup: Why's that necessary?

fago, line 56: hm?

src/Engine/RulesEventManager.php
```diff
@@ -0,0 +1,85 @@
...
+  public function processDefinition(&$definition, $plugin_id) {
+    parent::processDefinition($definition, $plugin_id);
+    if (!isset($definition['context'])) {
+      return;
+    }
+    // Convert the flat context arrays into ContextDefinition objects.
+    // @todo This code should be removed and we should pass this off to some
+    //   annotation reader code that converts plugin defintion parts into
+    //   objects.
+    foreach ($definition['context'] as $context_name => $values) {
```
fago, line 59: There is some definition fromArray() constructuror in Rules' context definition class now.

src/Entity/ReactionRule.php
```diff
@@ -0,0 +1,195 @@
...
+   * @var string
+   */
+  protected $tag = '';
+
+  /**
+   * The core version the Reaction rule was created for.
+   *
+   * @var int
+   */
+  protected $core = \Drupal::CORE_COMPATIBILITY;
```
fago, line 79: Why do we need that?

src/Entity/ReactionRule.php
```diff
@@ -0,0 +1,195 @@
...
+   * @var \Drupal\rules\Engine\ExpressionInterface
+   */
+  protected $expression;
+
+  /**
+   * The module implementing this Reaction rule.
+   *
+   * @var string
+   */
+  protected $module = 'rules';
```
fago, line 107: usually provider ?

src/EventSubscriber/GenericEventSubscriber.php
```diff
@@ -0,0 +1,91 @@
...
+    // Register this listener for every event that is used by a reaction rule.
+    $events = [];
+    $callback = ['onRulesEvent', 100];
+
+    // If there is no state service there is nothing we can do here.
+    if (!\Drupal::hasService('state')) {
+      return [];
+    }
+
+    // Since we cannot access the reaction rule config storage here we have to
```
fago, line 49: Why can't it be used here?

src/EventSubscriber/GenericEventSubscriber.php
```diff
@@ -0,0 +1,91 @@
...
+   * Reacts on the given event and invokes configured reaction rules.
+   *
+   * @param \Symfony\Component\EventDispatcher\GenericEvent $event
+   *   The event object containing context for the event.
+   * @param string $event_name
+   *   The event name.
+   */
+  public function onRulesEvent(GenericEvent $event, $event_name) {
+    // Load reaction rule config entities by $event_name.
+    $storage = $this->entityManager->getStorage('rules_reaction_rule');
```
fago, line 74: This really needs caching as in d7. Maybe let'S create a ticket for implementing it?

tests/src/Integration/Event/UserLoginTest.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Checks that the event "rules_user_login" is correctly defined.
+ */
+class UserLoginTest extends EventTestBase {
+
+  /**
+   * Tests the event metadata.
+   */
+  public function testUserLoginEvent() {
+    $event = $this->eventManager->createInstance('rules_user_login');
```
fago, line 19: Not sure there is much value in that. Rather than, maybe it makes sense to trigger the event with a real case and assert the event & defined contexts are there and accessible?

klausi, line 19: we have a real integration test also ... but yeah, not sure how we should unit test events on their own yet

config/schema/rules.schema.yml
klausi, line 56: That is basically a copy of View.php from core, which also saves the core compatibility flag. Not sure we need it, but I would keep it for now.

src/Engine/RulesEventManager.php
klausi, line 59: good catch, thanks!

src/Entity/ReactionRule.php
klausi, line 79: Because Views in core does it :-)

src/Entity/ReactionRule.php
klausi, line 107: Views also uses "module", so sticking with that for now.

src/EventSubscriber/GenericEventSubscriber.php
klausi, line 49: Will add the comment "This static method could be called early when the container is built, so the state service might no always be available."

src/EventSubscriber/GenericEventSubscriber.php
klausi, line 74: Done: https://www.drupal.org/node/2513076

PR #197, opened 2015-06-07
Separate the actions API from the core actions API.

fago:
Straight forward separation of the actions API. For now I've just taken over the existing access() method - it needs to be fixed though, for which we have a separate issue.


dasjo:
Needed rebasing.
https://github.com/fago/rules/pull/198

PR #199, opened 2015-06-19
Issue #2317223 Adding autosaving of the user account if it gets block…

karolkasas:
Adding auto save based on @klausi comments https://www.drupal.org/node/2317223#comment-9841589


klausi:
Merged, thanks!

PR #209, opened 2015-07-24
Port event_type is viewed event to D8

JM79:
https://www.drupal.org/node/2539112
(Apologies if I've pulled/pushed this incorrectly).


dasjo:
merged, thanks!

PR #210, opened 2015-07-25
Port rules init system event to D8

stephenpurkiss:
Hi,

I've had a quick go at starting on this - I've copied from what's been done so far but not entirely sure I got it right. From what I gather we're exposing symfony event through Drupal so perhaps this needs to link to the kernel.request event listed in rules.rules.events.yml, but I'm not sure how to do that...!


bojanz:
hook_init() doesn't actually exist anymore: https://www.drupal.org/node/2013014

klausi:
Yes, it should be enough to specify the event in the rules.rules_events.yml file since that hook is an event already :-)

stephenpurkiss:
Thanks, I kinda had a feeling that might be the case but as it was listed as a "to-do", I did. Feel free to close this and destroy the code in a mighty flash!

stephenpurkiss:
Oh look, there's a close button there, I have the power!

PR #211, opened 2015-07-28
Feature/2471465  port js css

mark-casias:
Pretty straight forward pull from D7 to D8, with a few formatting things that PHPStorm was complaining about. This is a start, though I am sure as we progress in getting the UI up, we'll have to make some changes.

https://www.drupal.org/node/2541766


klausi:
So you are just adding the CSS files but never including them? Am I missing something?

mark-casias:
Added some base form adjustments and attached the css / js. Created a libraries yaml file.

PR #212, opened 2015-08-01
Travis CI: Cache composer downloads, retry composer installs in case of failures

klausi:


klausi:
This has been merged.

PR #213, opened 2015-08-04
Fix nitpicks

bojanz:
I've been reading the Rules code tonight, came across some silly issues, decided to fix them.

Remarks:
- The config entities should have matching interfaces
- It's weird for entities to define links provided by the rules_ui module. Maybe it's time to kill the UI separation?


fubhy:
> It's weird for entities to define links provided by the rules_ui module. Maybe it's time to kill the UI separation?

Absolutely. We spoke about that a while ago and IIRC there was no argument for keeping the UI module separate apart from other priorities atm. ... But I am all for it.

src/Entity/ReactionRule.php
```diff
@@ -135,9 +135,6 @@ public function getExpression() {
   /**
    * Returns the Rules expression manager.
    *
-   * @todo Actually we should use dependency injection here, but is that even
-   *   possible with config entities? How?
```
klausi, line 139: Hm, should we remove that todo? We are using \Drupal here, which we should not do, so we should document why we do it.

klausi:
Merged everything except the todos, thanks!

bojanz:
Well, there's no way not to do it, D8 doesn't support DI in entities, typed data, a few other places. Rules is doing the same thing core is doing. Anyway, thanks for the merge.

PR #216, opened 2015-08-09
Fix configuration import with context config

fago:


klausi:
Looks good, but we should use ExpressionManagerInterface everywhere: type hints, docs, mock objects ...

fago:
thanks, yeah not sure the interface wasn't in there. I fixed that + fixed conflicts + merged to avoid more conflicts again.

PR #217, opened 2015-08-09
Port mail_to_users_of_role action to D8

stephenpurkiss:
Following up from the work done on https://github.com/fago/rules/pull/173 - I rebased onto the current rules 8.x-3.x, ran the tests and they were fine (PR173 suggests they still break but they don't). Also made one correction where it hadn't been changed from 'Action' to 'RulesAction'. Issue on d.o at https://www.drupal.org/node/2409059


src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Provides a 'Mail to users of a role' action.
+ *
+ * @RulesAction(
+ *   id = "rules_mail_to_users_of_role",
+ *   label = @Translation("Mail to users of a role"),
+ *   category = @Translation("System"),
+ *   context = {
+ *     "roles" = @ContextDefinition("entity",
```
klausi, line 26: This should of type "entity:user_role"

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory.
+   * @param \Drupal\user\UserStorage $userStorage
+   *   The user storage service.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, LoggerInterface $logger, MailManagerInterface $mail_manager, ConfigFactoryInterface $config_factory, UserStorage $userStorage) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->logger = $logger;
+    $this->mailManager = $mail_manager;
+    $this->config = $config_factory->get('site.config');
```
klausi: we should get the site.config in the create() method to not have the whole config factory as dependency here.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+      $container->get('plugin.manager.mail'),
+      $container->get('config.factory'),
+      $container->get('entity.manager')->getStorage('user')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
```
klausi: The summary method should be removed, this is now auto generated from the label in the plugin annotation.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+    return $this->t('Sends an e-mail to the users of a role');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $roles = $this->getContextValue('roles');
+
+    if (empty($roles)) {
```
klausi, line 118: This should not be necessary, the roles context is required, so it must always be there.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+      return $role->id();
+    }, $roles));
+    if (empty($rids)) {
+      return;
+    }
+
+    // Get now all the users that match the roles (at least one of the role).
+    $accounts = $this->userStorage
+      ->loadByProperties(['roles' => $roles]);
+    // @todo: Should we implement support for user tokens in subject and body? In the
```
klausi, line 133: This todo should be removed. We handle token replacements outside actions, so we are good here.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,169 @@
...
+    // Drupal 7 version it is not implemented.
+    $params = array(
+      'subject' => $this->getContextValue('subject'),
+      'body' => $this->getContextValue('body'),
+    );
+    $from = $this->getContextValue('from');
+    if (empty($from)) {
+      $from = $this->config->get('mail');
+    }
+    $key = $this->getPluginId();
```
klausi, line 143: Hm, the key should be something longer since the action can be used in multiple rules. Not sure how we should solve that, but please add a todo.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,238 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    $this->namespaces[] = '';
+
+    $this->logger = $this->getMock('Psr\Log\LoggerInterface');
```
klausi, line 63: Since Drupal now requires PHP 5.5 we can use LoggerInterface::class here instead of a raw string. Also elsewhere.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,238 @@
...
+   *
+   * @param string $call_number
+   *   The number of emails that should be sent.
+   * @param \Drupal\user\Entity\Role[] $roles
+   *   The array of Role objects to send the email to.
+   * @param \Drupal\user\Entity\User[] $users
+   *   The array of users that should get this email.
+   *
+   */
+  private function _testSendMailToRole($call_number, $roles, $users) {
```
klausi, line 190: do not start methods with underscore. Just don't start it with "test".

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,238 @@
...
+   */
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    $this->namespaces[] = '';
```
klausi, line 61: why is that line needed? Please add a comment.

stephenpurkiss:
Hmm, not sure about this as still seems I'm passing the factory through still which doesn't solve the original issue "we should get the site.config in the create() method to not have the whole config factory as dependency here." But I don't see how I can get just the site.config without the config factory. I may do tomorrow when I look into it more, just thought I'd post up what I had as it passed the tests ;)

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,163 @@
...
+    $this->mailManager = $mail_manager;
+    $this->config = $config;
+    $this->userStorage = $userStorage;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    $site_config = $container->get('config.factory');
```
klausi: this is not site_config, this should be $config_factory

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,163 @@
...
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('logger.factory'),
+      $container->get('plugin.manager.mail'),
+      $site_config,
```
klausi: so you are still passing the config factory right now.

super-romeo:
I want to move the issue forward.
I've started working.

klausi:
@superromeo yes, please create a new pull request.

super-romeo:
Sure, but I need do the work first :)

super-romeo:
Created new pull request here [#390](https://github.com/fago/rules/pull/390).

klausi:
continued in #390

PR #220, opened 2015-08-09
Port 'user_logout' event to D8

stephenpurkiss:
Adding user_logout event to rules.rules.events.yml, plus copying & adapting the user_login test to work for user_logout


klausi:
merged, thanks!

PR #222, opened 2015-08-10
Implemented allow NULL values on context defintions

klausi:
See https://www.drupal.org/node/2245039


src/Context/Annotation/ContextDefinition.php
```diff
@@ -38,6 +38,8 @@ public function __construct(array $values) {
     // Default to Rules context definition class.
     $values['class'] = isset($values['class']) ? $values['class'] : '\Drupal\rules\Context\ContextDefinition';
     parent::__construct($values);
+    $additional_values = $values + ['allow_null' => FALSE];
```
fago: That works, but could seems to be simpler if we'd just switch to the ContextDefinition::createFromArray(). Also, I guess it's better to centralize usual construction logic and bake class check etc. in there also.

src/Context/ContextDefinitionInterface.php
```diff
@@ -10,8 +10,33 @@
...
+ *
+ * The core interface is extended to add properties that are necessary for
+ * Rules.
  */
 interface ContextDefinitionInterface extends ContextDefinitionInterfaceCore {
 
+  /**
+   * Determines if the context value is allowed to be NULL.
+   *
+   * Usually Rules will not pass any NULL values as argument, but abort the
```
fago: Let's better keep docs general here as the interface is general and does not enforce usage by Rules also. We might be able to move improvements in core later etc., docs should be written to facilitate that imho.

src/Context/ContextHandlerTrait.php
```diff
@@ -46,6 +46,14 @@ protected function mapContext(CoreContextAwarePluginInterface $plugin, RulesStat
       // Check if a data selector is configured that maps to the state.
       if (isset($this->configuration['context_mapping'][$name])) {
         $typed_data = $state->applyDataSelector($this->configuration['context_mapping'][$name]);
+
+        if ($typed_data->getValue() === NULL && !$definition->isAllowedNull()) {
```
fago: It's not sufficient to check this for mapping only, we need to check it regardless where the data comes from.

src/Engine/RulesState.php
```diff
@@ -130,7 +130,14 @@ public function applyDataSelector($selector, $langcode = LanguageInterface::LANG
       // Drill down to the next step in the data selector.
       if ($typed_data instanceof ListInterface || $typed_data instanceof ComplexDataInterface) {
         try {
-          $typed_data = $typed_data->get($name);
+          $child_typed_data = $typed_data->get($name);
+
+          // If the list is empty the returned value can be NULL here, so we
+          // create a new empty item.
+          if ($child_typed_data === NULL) {
+            $child_typed_data = $typed_data->appendItem();
```
fago: uhm, why would we change data here?

src/Tests/NodeIntegrationTest.php
```diff
@@ -255,4 +256,84 @@ public function testDataSet() {
...
+
+    $node = $entity_manager->getStorage('node')
+      ->create([
+        'title' => 'test',
+        'type' => 'page',
+      ]);
+
+    // Configure a simple rule with one action. The node ID is NULL and does not
+    // exist yet.
+    $action = $this->expressionManager->createInstance('rules_action',
```
fago: should use createAction() on the expression manager. Ideally, this wouldb extended to take ContextConfig as optional second argument imho, but that's another issue.

src/Tests/NodeIntegrationTest.php
```diff
@@ -255,4 +256,84 @@ public function testDataSet() {
     $this->assertNotNull($node->id(), 'Node ID is set, which means that the node has been auto-saved.');
   }
 
+  /**
+   * Tests that a required context value that is NULL throws an exception.
+   */
+  public function testRequiredNullValue() {
```
fago: it seems wrong to have that in NodeIntegrationTests, can we move this to engine tests?

fago:
Thanks for the update, merged!

PR #223, opened 2015-08-21
Implemented parameter restriction on the context definition interface

klausi:
See https://www.drupal.org/node/2245039


src/Context/ContextDefinition.php
```diff
@@ -38,6 +39,15 @@ class ContextDefinition extends ContextDefinitionCore implements ContextDefiniti
   protected $allowNull = FALSE;
 
   /**
+   * The parameter restriction of this context.
+   *
+   * @see \Drupal\rules\Context\ContextDefinitionInterface::getParameterRestriction()
```
fago: just @see ::getParameterRestriction() would be enough

src/Context/ContextDefinitionInterface.php
```diff
@@ -18,6 +18,14 @@
 interface ContextDefinitionInterface extends ContextDefinitionInterfaceCore {
 
   /**
+   * Constants for the context parameter restriction mode.
+   *
+   * @see ::getParameterRestriction()
+   */
+  const PARAMETER_RESTRICTION_INPUT = 'input';
```
fago: I think the wording needs work. it's no parameter any more but context in d8 - so the name does not fit any more. context restriction is not really specific, so what about "assignment restriction" - setAssignmentRestriction('input') ?

src/Context/ContextDefinitionInterface.php
```diff
@@ -35,4 +43,26 @@ public function isAllowedNull();
...
+   * @return string|null
+   *   Either PARAMETER_RESTRICTION_INPUT for context parameters that are only
+   *   allowed to be provided as input values, PARAMETER_RESTRICTION_SELECTOR
+   *   for context parameters that must be provided as data selectors or NULL if
+   *   there is no restriction for this context.
+   */
+  public function getParameterRestriction();
+
+  /**
+   * Sete the parameter restriction mode for this context.
```
fago: sete

klausi:
Good idea with the "assignment restriction" naming, changed that.

klausi:
merged.

PR #230, opened 2015-09-20
Issue #2571501: When executed via a reaction rule, RulesConditions do…

torenware:
… not negate

When processing an event, conditions are called via executeWithState; this routine must apply negation.

Missing test; will need some help with that.


klausi:
A similar fix has been committed.

PR #232, opened 2015-09-21
Update DataConvert action to use doExecute() with context arguments

ndeet:
Issue on d.o: https://www.drupal.org/node/2572119

I made the doExecute() function protected as it is also protected in EntityFetchByField class (given as boilerplate on d.o. issue above).

Additionally fixed a boolean from "false" to "FALSE", hope it is ok to not make an extra pull request - let me know if I should.


ndeet:
Think this one can be closed as @borissondagh did it too but cleaner with docblock.
https://github.com/fago/rules/pull/240

klausi:
merged with #304

PR #233, opened 2015-09-22
Feature/events

fago:
Improve how events are defined.


klausi:
Meh, public properties. But trying to build a method name could also be ugly:

``` php
$name_parts = array_map('ucfirst', explode('_', $context_name));
$method = 'get' . implode('', $name_parts);
$value = $event->$method;
```

Not sure what would be best.

fago:
Yes, still I think supporting methods like this also would be good - so you can easily add some simple calculations etc. Whether methods can provide a solution for lazy-loading we'll have to figure out separately.

Howsoever, I think public properties are better than methods here if there is no need to add any logic. Thus, the event class basically is a simple class based replacemented for an array, so it's properly documented and has auto-completion. I think public properties are the best compromise combining the advantages of getter/setter methods (properly defined & documented!) and GenericEvent (ease of defining it, DX), thus should be the way to go. I don't see writability of them as problematic - first off I don't see that we have to prevent that, second GenericEvent has write support also.

klausi:
I find it a bit ugly that you have to write more boilerplate code that just duplicates the meta data info from rules.rules.events.yml.

But yeah, I don't mind too much.

PR #235, opened 2015-09-23
Issue #2572945: doExecute / doEvaluate for DataIsEmpty

torenware:
Drupal Issue: https://www.drupal.org/node/2572945

Implementation of the modified plugin, and the modified test.

<del>The test of complex data interfaces no longer works, and may not even be relevant anymore, since RuleConditionBase::evaluate() unwraps the data for us.  I've commented out that part of the test.</del>

To get the tests to work, the mocks needed to be changed so that the complex data object would not get automatically unwrapped.  The mocks verify that the object indeed gets passed into the condition, and behaves as expected.

### DIFF
--- a/src/Plugin/Condition/DataIsEmpty.php
+++ b/src/Plugin/Condition/DataIsEmpty.php
@@ -36,17 +36,10 @@ class DataIsEmpty extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
-  public function evaluate() {
-    $data = $this->getContext('data')->getContextData();
+  protected function doEvaluate($data) {
     if ($data instanceof ComplexDataInterface || $data instanceof ListInterface) {
       return $data->isEmpty();
     }
-    $value = $data->getValue();
-    // For some primitives we can rely on PHP's type casting to boolean.
-    if ($data instanceof StringInterface || $data instanceof IntegerInterface || $data instanceof BooleanInterface) {
-      return !isset($value) || !$value;
-    }
-    return !isset($value);
+    return empty($data);
   }
-
 }
--- a/tests/src/Integration/Condition/DataIsEmptyTest.php
+++ b/tests/src/Integration/Condition/DataIsEmptyTest.php
@@ -9,7 +9,10 @@ namespace Drupal\Tests\rules\Integration\Condition;
 
 use Drupal\Core\Plugin\Context\Context;
 use Drupal\Core\TypedData\ComplexDataInterface;
+use Drupal\Core\TypedData\DataDefinitionInterface;
+use Drupal\Core\TypedData\TypedDataManagerInterface;
 use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
+use Prophecy\Argument;
 
 /**
  * @coversDefaultClass \Drupal\rules\Plugin\Condition\DataIsEmpty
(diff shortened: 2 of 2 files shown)


torenware:
This is failing due to https://www.drupal.org/node/2573007, PR https://github.com/fago/rules/pull/234.  Commit that, and this will pass its tests.

tests/src/Integration/Condition/DataIsEmptyTest.php
```diff
@@ -39,17 +39,17 @@ public function setUp() {
...
-
-    $this->condition->getContext('data')->setContextData($entity_adapter_empty->reveal());
-    $this->assertTrue($this->condition->evaluate());
-
-    $entity_adapter_full = $this->prophesize(ComplexDataInterface::class);
-    $entity_adapter_full->isEmpty()->willReturn(FALSE)->shouldBeCalledTimes(1);
-
-    $this->condition->getContext('data')->setContextData($entity_adapter_full->reveal());
-    $this->assertFalse($this->condition->evaluate());
+    //$entity_adapter_empty = $this->prophesize(ComplexDataInterface::class);
```
fago: ?

src/Plugin/Condition/DataIsEmpty.php
```diff
@@ -36,7 +36,7 @@ class DataIsEmpty extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
-  public function evaluate() {
+  public function _evaluate() {
```
fago: Please remove this.

tests/src/Integration/Condition/DataIsEmptyTest.php
torenware: Fago,

The problem here is that the mocks for $entity_adapter_empty and $entity_adapter_full are both now invalid.  phpunit puts out errors like these:

1) Drupal\Tests\rules\Integration\Condition\DataIsEmptyTest::testConditionEvaluation
Prophecy\Exception\Call\UnexpectedCallException: Method call:
- getDataDefinition()
  on Double\ComplexDataInterface\P5 was not expected, expected calls were:
- isEmpty()

If you like, I can look up how to mock getDataDefinition(), but it's not trivial, and there are not currently any examples in the Rules codebase.  Note also that we use Prophecy, while Core uses the standard mock builder, so there are no relevant examples in core either.

So in any case, the mocks have to be changed in some way or the other.

src/Plugin/Condition/DataIsEmpty.php
torenware: Hmm, missed that.  Will do.

tests/src/Integration/Condition/DataIsEmptyTest.php
torenware: What exactly are these mocks intended to test?

As I said before, the mocks need to change.  The following works, but I don't know if this tests what you really want tested:

``` php
    // Test a ComplexDataInterface object.
    $data_definition_interface = $this->prophesize(DataDefinitionInterface::class);
    $data_definition_interface->getDataType()->willReturn('any');

    $entity_adapter_empty = $this->prophesize(ComplexDataInterface::class);
    $entity_adapter_empty->getDataDefinition()->willReturn($data_definition_interface);
    $entity_adapter_empty->getValue()->willReturn(FALSE);

    $this->condition->getContext('data')->setContextData($entity_adapter_empty->reveal());
    $this->assertTrue($this->condition->evaluate());

    $entity_adapter_full = $this->prophesize(ComplexDataInterface::class);
    $entity_adapter_full->getDataDefinition()->willReturn($data_definition_interface);
    $entity_adapter_full->getValue()->willReturn(TRUE);

    $this->condition->getContext('data')->setContextData($entity_adapter_full->reveal());
    $this->assertFalse($this->condition->evaluate());
```

`isEmpty()` has to go, because it really is never called:  by the time doEvaluate() sees the complex data, it's already been unwrapped to the data itself.  The call to `getValue()` has already done the conversion.

What do you you guys want me to do here?

torenware:
Fago -- stage 2 of the process.  I've removed the old code (oops!) and done what I could to fix the mocks.  The old mocks can't be used exactly as written; the sequence of calls they describe won't happen for a ComplexDataInterface object.  I may need Klausi's help to figure out what the mocks need to be, and will add any changes he suggests.

tests/src/Integration/Condition/DataIsEmptyTest.php
```diff
@@ -39,15 +40,20 @@ public function setUp() {
    */
   public function testConditionEvaluation() {
     // Test a ComplexDataInterface object.
+    $data_definition_interface = $this->prophesize(DataDefinitionInterface::class);
+    $data_definition_interface->getDataType()->willReturn('any');
+    
     $entity_adapter_empty = $this->prophesize(ComplexDataInterface::class);
-    $entity_adapter_empty->isEmpty()->willReturn(TRUE)->shouldBeCalledTimes(1);
-
+    $entity_adapter_empty->getDataDefinition()->willReturn($data_definition_interface);
```
klausi: Looks like you forgot to call ->reveal() on $data_definition_interface here?

tests/src/Integration/Condition/DataIsEmptyTest.php
```diff
@@ -39,15 +40,20 @@ public function setUp() {
...
+    $entity_adapter_empty->getValue()->willReturn(FALSE);
+        
     $this->condition->getContext('data')->setContextData($entity_adapter_empty->reveal());
     $this->assertTrue($this->condition->evaluate());
-
+    
     $entity_adapter_full = $this->prophesize(ComplexDataInterface::class);
-    $entity_adapter_full->isEmpty()->willReturn(FALSE)->shouldBeCalledTimes(1);
-
+    $entity_adapter_full->getDataDefinition()->willReturn($data_definition_interface);
```
klausi: same here?

torenware:
The port was easy; the problem was how to get the test to actually cover the ComplexDataType part of the test.  The mock's kind of evil, but it works, and the mocked data type does indeed cover the relevant path in the plugin.

src/Plugin/Condition/DataIsEmpty.php
```diff
@@ -36,17 +36,10 @@ class DataIsEmpty extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
-  public function evaluate() {
-    $data = $this->getContext('data')->getContextData();
+  protected function doEvaluate($data) {
     if ($data instanceof ComplexDataInterface || $data instanceof ListInterface) {
       return $data->isEmpty();
```
klausi, line 41: So this is now dead code and will never be executed, right? $data will never conform to those interfaces because $data is not typed data here?

torenware, line 41: @klausi Not exactly.  The problem is that most implementations of ComplexDataInterface are configured to auto-unwrap; see the implementation of TypedDataManager::getCanonicalRepresentation(), which checks for a 'unwrap_for_canonical_representation' flag in a typed data definition. All common types I've looked at have this definition property TRUE.  So typically, the data unwraps.

Implementations of ContentEntity, however, do not unwrap, but there are places in the code that guarantee that the isEmpty() will always be false.  Entities want pretty badly to be "not empty".

So to get to that line of code, you need a complex data instance that (1) does not unwrap, and (2) can actually return isEmpty() as TRUE.  As you can see, you need to do a weird mock to get that.  I'm not sure of a practical use case for the combination, but the interfaces definitely allow for this.  So the code is not dead. A content entity will hit this line, and will return FALSE.

A complex data object that is both non-unwrappable and empty, however, is a weird edge case.

fago:
see https://www.drupal.org/node/2572945#comment-10657040

PR #236, opened 2015-09-23
Issue #2573045: doEvaluate for rules_data_comparison

torenware:
Drupal issue https://www.drupal.org/node/2573045

<del>I needed to swap the order of the context items, since it appears that PHP gets very annoyed when an optional argument is not at the end of the list.  In particular: if you pass the non-final argument as NULL, call_func_array does strange things.

Test needed to be tweaked as well; it seems to matter what order setContextValue() is called, so 'operator' is now always last.</del>

Current Rules HEAD now does the right thing with preserving annotation order.

### DIFF
--- a/src/Plugin/Condition/DataComparison.php
+++ b/src/Plugin/Condition/DataComparison.php
@@ -24,6 +24,7 @@ use Drupal\rules\Core\RulesConditionBase;
  *     "operator" = @ContextDefinition("string",
  *       label = @Translation("Operator"),
  *       description = @Translation("The comparison operator."),
+ *       default_value = "==",
  *       required = FALSE
  *     ),
  *     "value" = @ContextDefinition("any",
@@ -39,13 +40,25 @@ use Drupal\rules\Core\RulesConditionBase;
 class DataComparison extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Evaluate the data comparison.
+   *
+   * @param mixed $data
+   *   Supplied data to test.
+   * @param string $operator
+   *   Data comparison operator. Typically one of:
+   *     - "=="
+   *     - "<"
+   *     - ">"
+   *     - "contains" (for strings or arrays)
+   *     - "IN" (for arrays or lists).
+   * @param mixed $value
+   *   The value to be compared against $data.
+   *
+   * @return bool
+   *   The evaluation of the condition.
    */
-  public function evaluate() {
-    $data = $this->getContextValue('data');
-    $operator = $this->getContext('operator')->getContextData() ? $this->getContextValue('operator') : '==';
-    $value = $this->getContextValue('value');
-
+  protected function doEvaluate($data, $operator, $value) {
+    $operator = $operator ? $operator : '==';
     switch ($operator) {
       case '<':
         return $data < $value;


src/Plugin/Condition/DataComparison.php
```diff
@@ -21,14 +21,14 @@
...
  *       description = @Translation("The data to be checked to be empty, specified by using a data selector, e.g. 'node:uid:entity:name:value'.")
  *     ),
+ *     "value" = @ContextDefinition("any",
+ *        label = @Translation("Data value"),
+ *        description = @Translation("The value to compare the data with.")
+ *     ),
  *     "operator" = @ContextDefinition("string",
  *       label = @Translation("Operator"),
  *       description = @Translation("The comparison operator."),
  *       required = FALSE
```
fago, line 28: We must keep the operator second, as we'll need it that way in future (operator plugins that influence what you can select after). Let's better make it required instead - it should be really, but have a default value! NULL is not a valid value for it. Might be we have to adapt tests also for that.

torenware, line 28: This turns out be a very subtle bug.  Going through the debugger, I'm discovering that when RulesConditionBase::evaluate() calls $this->getContexts(), the ordering is NOT according to the order as defined in the annotation.  It's according to the order that the plugin is notified about the contexts at run time.  So you're not seeing annotation order; you're seeing the order in which $this->context was written to.  So order can get swapped.

If you want to make sure that you're getting things in annotation order, then you need to fill in the array you pass back really _is_ the order given back by $this->getContextDefinitions().  The function needs to be rewritten as

drupal/core/lib/Drupal/Component/Plugin/ContextAwarePluginBase.php, line 79:

``` php
  public function getContexts() {
    // Make sure all context objects are initialized.
    $ordered_contexts = [];
    foreach ($this->getContextDefinitions() as $name => $definition) {
      $ordered_contexts[$name] = $this->getContext($name);
    }
    return $ordered_contexts;
  }
```

This file is in Core, and needs to be reported against Core.  For our purposes, we should probably just override the function in our plugin base class.

But if I make this change, the old tests work absolutely fine.

torenware:
Original tests should pass once https://www.drupal.org/node/2573823 is committed.

src/Plugin/Condition/DataComparison.php
```diff
@@ -39,13 +40,10 @@
 class DataComparison extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Evaluate the data comparison
```
nlisgo: Needs a full stop at the end of the line.

src/Plugin/Condition/DataComparison.php
```diff
@@ -67,6 +65,7 @@ public function evaluate() {
         }
         return $data == $value;
     }
+
```
nlisgo: Remove newline.

src/Plugin/Condition/DataComparison.php
```diff
@@ -67,6 +65,7 @@ public function evaluate() {
         }
         return $data == $value;
     }
+    
   }
-
+
```
nlisgo: Remove whitespace.

torenware:
Slight formatting fix.  Should run green against the current 8.x-3.x.

The original ordering of the items has been restored, per @fago

torenware:
We now pass the original tests.

src/Plugin/Condition/DataComparison.php
```diff
@@ -39,13 +40,10 @@
 class DataComparison extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Evaluate the data comparison.
    */
```
klausi, line 59: Can we have param docs for this?

torenware:
@klausi: I've added param documentation, and made sure phpcs is happy happy.  Should be good.

src/Plugin/Condition/DataComparison.php
```diff
@@ -24,7 +24,8 @@
  *     "operator" = @ContextDefinition("string",
  *       label = @Translation("Operator"),
  *       description = @Translation("The comparison operator."),
- *       required = FALSE
+ *       default_value = "==",
+ *       required = TRUE
```
klausi: why is it now required? since we have a default value now it should not be required?

torenware: @klausi -- I think that this was done due to <del>[Context in Rules plugins does not respect annotation order](https://www.drupal.org/node/2573823)</del>.  Until that issue was fixed, unless it was marked "required",  `required = TRUE` was needed for the default value to get used, oddly enough.

I've just done some tests, and this is no longer an issue.  I'll change this back to `required = FALSE`.

torenware:
Annotation's been changed back.  Newer annotation processing code doesn't need `required = TRUE` anymore.

klausi:
merged, thanks!

PR #237, opened 2015-09-24
Issue #2573823: Context in Rules plugins does not respect annotation order

torenware:
Drupal issue:  https://www.drupal.org/node/2573823

Here's a work-around that works in Rules.

Note: I only saw fago's comment in the issue queue after I finished this.  So my solution will be a little different than what he suggested there. 

It does run all of the existing tests for rules, and also works with DataComparisonTest.


torenware:
I've made sure this applies clean to today's build, and run phpcs on the files.

klausi:
committed a different approach, thanks.

PR #238, opened 2015-09-25
Issue #2574513: doEvaluate for rules_list_contains

torenware:
Port of class DataListContains. Tests run green w/o modification.

Drupal issue: https://www.drupal.org/node/2574513

### DIFF
--- a/src/Plugin/Condition/DataListContains.php
+++ b/src/Plugin/Condition/DataListContains.php
@@ -37,12 +37,14 @@ use Drupal\Core\Entity\EntityInterface;
 class DataListContains extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Evaluate whether the list has the item.
+   *
+   * @param array|ListInterface $list
+   *   List to be searched.
+   * @param mixed $item
+   *   Item to be found in list.
    */
-  public function evaluate() {
-    $list = $this->getContextValue('list');
-    $item = $this->getContextValue('item');
-
+  protected function doEvaluate($list, $item) {
     if ($item instanceof EntityInterface && $id = $item->id()) {
       // Check for equal items using the identifier if there is one.
       foreach ($list as $list_item) {


src/Plugin/Condition/DataListContains.php
```diff
@@ -35,14 +35,11 @@
  * @todo: Add info alter
  */
 class DataListContains extends RulesConditionBase {
-
+
```
klausi: trailing space

src/Plugin/Condition/DataListContains.php
```diff
@@ -35,14 +35,11 @@
  * @todo: Add info alter
  */
 class DataListContains extends RulesConditionBase {
-
+  
   /**
-   * {@inheritdoc}
+   * Evaluate whether the list has the item
    */
```
klausi, line 46: Please document the new params

src/Plugin/Condition/DataListContains.php
```diff
@@ -52,8 +49,7 @@ public function evaluate() {
       }
       return FALSE;
     }
-
-    return in_array($item, $list);
+    return in_array($item, $list);
```
klausi: Unrelated change that should be removed?

src/Plugin/Condition/DataListContains.php
torenware, line 46: I'm declaring $list as an array, since data type 'list' stores its data internally as an array.  Not sure in general how to translate some TypedData types into PHP types for this purpose.

src/Plugin/Condition/DataListContains.php
torenware: This looks like an area that varies a bit from plugin to plugin.  Similar plugins appear to put return statements flush with the "}" of the previous line.  This is not universal, but most plugins that have this kind of code structure omit the extra line.

If we want to enforce this, I guess we could always use a grunt plugin; must be one for this.

torenware:
I've cleaned this up with drupalcs; param $list has been redeclared to better match the 'list' data type.

klausi:
merged, thanks!

PR #240, opened 2015-09-25
Issue #2472815 doExecute for DataConvert

botris:
Updated the execute() function to doExecute() for the DataConvert class.
Issue #2472815

### DIFF
--- a/src/Plugin/RulesAction/DataConvert.php
+++ b/src/Plugin/RulesAction/DataConvert.php
@@ -43,13 +43,15 @@ class DataConvert extends RulesActionBase {
 
   /**
    * Executes the plugin.
+   *
+   * @param $value
+   *   The input value
+   * @param string $target_type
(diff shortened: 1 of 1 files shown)


src/Plugin/RulesAction/DataConvert.php
```diff
@@ -43,13 +43,15 @@ class DataConvert extends RulesActionBase {
 
   /**
    * Executes the plugin.
+   *
+   * @param $value
```
klausi, line 47: data type missing, see https://www.drupal.org/coding-standards/docs#param

klausi, line 47: If it can be anything we can use "mixed"

ndeet:
I took over, we should close this. https://github.com/fago/rules/pull/304

klausi:
closing.

PR #243, opened 2015-09-25
Updating UserBlock to use doExecute function

ndeet:


src/Plugin/RulesAction/UserBlock.php
```diff
@@ -75,14 +76,12 @@ public static function create(ContainerInterface $container, array $configuratio
   protected $saveLater = FALSE;
 
   /**
-   * {@inheritdoc}
+   * Block a user.
+   *
+   * @param UserInterface
```
klausi: param name is missing. And the type should be fully namespaced

klausi:
merged, thanks

PR #171, opened 2015-04-16
Settings page

milkovsky:

PR #244, opened 2015-09-25
Issue #2471693: Port Rules base settings

morvans:
New PR based on work in https://github.com/milkovsky/rules/tree/settings-2471693
Some conflicts needed to be resolved.
It also removes useless RulesUiController

d.o issue : https://www.drupal.org/node/2471693


morvans:
It basically replaces https://github.com/fago/rules/pull/171

dasjo:
It seems like this branch could use some clean-up and maybe also contains code changes which aren't related to the issue itself

morvans:
I've squashed this down to 2 commits (settings schema changes & provide config form)
Not sure if we really need test case for the form itself

src/Controller/RulesController.php
```diff
@@ -13,16 +13,5 @@
  * Returns responses for Rules routes.
  */
 class RulesController extends ControllerBase {
```
klausi: So this class is not used anymore? Can we remove it?

src/Form/RulesSettingsForm.php
```diff
@@ -0,0 +1,76 @@
...
+      '#type' => 'radios',
+      '#title' => t('Log level'),
+      '#options' => array(
+        LogLevel::WARNING => t('Log all warnings and errors'),
+        LogLevel::ERROR => t('Log errors only'),
+      ),
+      '#default_value' => $config->get('log_level') ? $config->get('log_level') : LogLevel::WARNING,
+      '#description' => t('Evaluations errors are logged to available loggers.'),
+      '#states' => array(
+        // Hide the regions settings when the debug log is disabled.
```
klausi: Region settings? You mean hiding the log level settings?

src/Controller/RulesController.php
morvans: You were more or less proposing to keep it even if it's now empty. I've no strong opinion on this one. For now, we can remove it. What do you think ?

src/Form/RulesSettingsForm.php
morvans: I think it was what the original author meant. I'll fix that one.

src/Controller/RulesController.php
klausi: yes, remove because it is unused

src/Form/RulesSettingsForm.php
```diff
@@ -0,0 +1,76 @@
...
+    );
+    $form['log_level'] = array(
+      '#type' => 'radios',
+      '#title' => t('Log level'),
+      '#options' => array(
+        LogLevel::WARNING => t('Log all warnings and errors'),
+        LogLevel::ERROR => t('Log errors only'),
+      ),
+      '#default_value' => $config->get('log_level') ? $config->get('log_level') : LogLevel::WARNING,
+      '#description' => t('Evaluations errors are logged to available loggers.'),
```
klausi: we must not call t() directly from within a class. There should be $this->t().

morvans:
> maybe also contains code changes which aren't related to the issue itself

I Agree. Initially, I basically just rebased work from another PR. What do you suggest ? Splitting in multiple issues ?

src/Form/RulesSettingsForm.php
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('rules.settings');
+
+    $form['log'] = array(
+      '#type' => 'checkbox',
+      '#title' => $this->t('Log debug information'),
```
klausi, line 41: Can we use the description from the Drupal 7 version? "Log debug information to the system log". Rules will always be logging, it is just a question whether it will write that to the system log or not.

config/install/rules.settings.yml
```diff
@@ -1,4 +1,2 @@
-log_errors: warning
-debug_log: false
-debug: false
-log_level: info
+log: false
+log_level: warning
```
klausi, line 2: I think "debug_log" should stay as "debug_log". Let's rename "debug" to "debug_screen" or "debug_message" or "debug_html" to make it clear that it prints debug info to the current page. I like "debug_screen" most.

klausi:
continued in #366

PR #245, opened 2015-09-25
Updating EntityDelete action to use doExecute()

ndeet:


klausi:
merged, thanks.

PR #246, opened 2015-09-25
applied ContexDefinition::create() instead of many constructors arguments

diego75l27:
https://www.drupal.org/node/2305573


src/Plugin/RulesAction/EntityCreateDeriver.php
```diff
@@ -67,7 +67,9 @@ public function getDerivativeDefinitions($base_plugin_definition) {
         'entity_type_id' => $entity_type_id,
         'context' => [],
         'provides' => [
-          'entity' => new ContextDefinition("entity:$entity_type_id", $entity_type->getLabel()),
+          'entity' => ContextDefinition::create("entity:$entity_type_id")
+            ->setLabel(t($entity_type->getLabel()))
```
klausi: Don't use t() here, that's why the tests break.

klausi:
merged

PR #248, opened 2015-09-25
Updating SystemSendMail action to use doExecute()

ndeet:


src/Plugin/RulesAction/SystemSendEmail.php
```diff
@@ -99,16 +97,24 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   * @param string|null $reply
+   *    (optional) Reply to email address.
+   * @param \Drupal\Core\Language\LanguageInterface|null $language
+   *    (optional) Language code.
    */
-  public function execute() {
-    $to = $this->getContextValue('to');
-    $reply = $this->getContextValue('reply');
-    $language = $this->getContextValue('language');
+  protected function doExecute($to, $subject, $message, $reply, LanguageInterface $language = NULL) {
```
klausi: So we should have a default value of NULL for $reply, right?

src/Plugin/RulesAction/SystemSendEmail.php
```diff
@@ -99,16 +97,24 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Send a system email.
+   *
+   * @param array $to
```
klausi: Let's use "string[]" as type hint, see https://www.drupal.org/coding-standards/docs#types

src/Plugin/RulesAction/SystemSendEmail.php
```diff
@@ -99,16 +97,24 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   *    Email addresses of the recipients.
+   * @param string $subject
+   *    Subject of the email.
```
klausi: 3 spaces instead of 2

klausi:
merged

PR #249, opened 2015-09-25
Updating SendAccountMail action to use doExecute()

ndeet:


src/Plugin/RulesAction/SendAccountEmail.php
```diff
@@ -33,12 +34,15 @@
 class SendAccountEmail extends RulesActionBase {
 
   /**
-   * {@inheritdoc}
+   * Send account email.
+   *
+   * @param \Drupal\user\UserInterface $account
+   *   User who should receive the notification.
+   * @param string $operation_type
```
klausi: The variable name should always match the context name we defined to avoid confusion. Thinking about it, I like "email_type" more since it is in fact the type of email that is sent out. It is confusing that _user_mail_notify() call it $op, but I think our naming is better. So let's stick to $email_type.

ndeet:
done

klausi:
merged

PR #251, opened 2015-09-25
Updating UserRoleAdd to use doExecute()

ndeet:
https://www.drupal.org/node/2572119


src/Plugin/RulesAction/UserRoleAdd.php
```diff
@@ -40,11 +41,14 @@ class UserRoleAdd extends RulesActionBase {
   protected $saveLater = FALSE;
 
   /**
-   * {@inheritdoc}
+   * Assign role to a user.
+   *
+   * @param \Drupal\user\UserInterface $account
+   *   User object.
+   * @param array $roles
```
klausi: This should use the "\Drupal\user\RoleInterface[]" data type instead of "array".

src/Plugin/RulesAction/UserRoleAdd.php
```diff
@@ -40,11 +41,14 @@ class UserRoleAdd extends RulesActionBase {
...
+   *   Array of UserRoles to assign.
    */
-  public function execute() {
-    $account = $this->getContextValue('user');
-    $roles = $this->getContextValue('roles');
+  protected function doExecute(UserInterface $account, $roles) {
```
klausi: $roles should be type-hinted as array

ndeet:
Right, thanks, fixed.

klausi:
merged, thanks!

PR #253, opened 2015-09-26
Updating PathAliasCreate to use doExecute().

ndeet:


src/Plugin/RulesAction/PathAliasCreate.php
```diff
@@ -78,12 +78,16 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Creates an alias for a exsisting path.
```
klausi: "for an existing path"

ndeet:
Fixed, seems I have an ex(s)isting writing problem.

klausi:
merged, thanks!

PR #257, opened 2015-09-29
Issue #2575977: remove use of now protected Context::setContextValue(…

torenware:
…) from Rules.

This is a first pass; Kernel tests setting NULL contexts still fail.

Issue https://www.drupal.org/node/2575977


klausi:
Thanks, fixed this in a slightly different way.

PR #259, opened 2015-10-08
#2572119 Changing execute to doExecute in VariableAdd.

emclaughlin1215:

### DIFF
--- a/src/Plugin/RulesAction/VariableAdd.php
+++ b/src/Plugin/RulesAction/VariableAdd.php
@@ -29,9 +29,12 @@ use Drupal\rules\Core\RulesActionBase;
 class VariableAdd extends RulesActionBase {
 
   /**
-   * Executes the plugin.
+   * Add a variable.
+   *
+   * @param string $value
+   *   The variable to add.
    */
-  public function execute() {
-    $this->setProvidedValue('variable_added', $this->getContext('value')->getContextValue());
+  public function doExecute($value) {
+    $this->setProvidedValue('variable_added', $value);
   }
 }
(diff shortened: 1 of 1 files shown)


emclaughlin1215:
Test are failing because of a protected method in the core Context.php class (setContextValue). I'm not sure what to do with that information.

klausi:
yep, we are currently working at fixing the tests in #258

torenware:
Pulled @emclaughlin1215's patch and ran it against today's Rules + Core with @klausi's Context patch.  The tests run green by me.

src/Plugin/RulesAction/VariableAdd.php
```diff
@@ -29,9 +29,12 @@
 class VariableAdd extends RulesActionBase {
 
   /**
-   * Executes the plugin.
+   * Add a variable.
+   *
+   * @param string $value
```
klausi: The value can be of any type, so the data type should be "mixed" here.

emclaughlin1215:

src/Plugin/RulesAction/VariableAdd.php
```diff
@@ -29,9 +29,12 @@
...
   /**
-   * Executes the plugin.
+   * Add a variable.
+   *
+   * @param mixed $value
+   *   The variable to add.
    */
-  public function execute() {
-    $this->setProvidedValue('variable_added', $this->getContext('value')->getContextValue());
+  public function doExecute($value) {
```
klausi, line 37: should be protected, not public.

ndf:
Moved to https://github.com/fago/rules/pull/305

klausi:
closing

PR #260, opened 2015-10-09
Issue #2585019: Implementing doExecute() for the DataListItemRemove a…

emclaughlin1215:
…ction.

See d.o issue: https://www.drupal.org/node/2585019 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/RulesAction/DataListItemRemove.php
+++ b/src/Plugin/RulesAction/DataListItemRemove.php
@@ -33,7 +33,7 @@ use Drupal\rules\Core\RulesActionBase;
 class DataListItemRemove extends RulesActionBase {
 
   /**
-   * Executes the Plugin.
+   * Removes an item from a list.
    *
    * @param array $list
    *   An array to remove an item from.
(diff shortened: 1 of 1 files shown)


src/Plugin/RulesAction/DataListItemRemove.php
```diff
@@ -33,12 +33,14 @@
 class DataListItemRemove extends RulesActionBase {
 
   /**
-   * {@inheritdoc}
+   * Executes the Plugin.
```
klausi: This should be a short description what the action does.

emclaughlin1215:

src/Plugin/RulesAction/DataListItemRemove.php
```diff
@@ -33,12 +33,14 @@
...
+   * @param array $list
+   *   An array to remove an item from.
+   * @param mixed $item
+   *   An item to remove from the array.
    */
-  public function execute() {
-    $list = $this->getContextValue('list');
-    $item = $this->getContextValue('item');
-
+  public function doExecute($list, $item) {
```
klausi, line 43: should be protected, not public.

klausi:
continued in #300

PR #263, opened 2015-10-09
Issue #2585403: Implementing doEvaluate() for PathHasAlias.

emclaughlin1215:
See d.o issues: https://www.drupal.org/node/2585403 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/PathHasAlias.php
+++ b/src/Plugin/Condition/PathHasAlias.php
@@ -77,14 +77,14 @@ class PathHasAlias extends RulesConditionBase implements ContainerFactoryPluginI
    *
    * @param string $path
    *   The path to check.
-   * @param mixed $language
+   * @param \Drupal\Core\Language\LanguageInterface|null $language
    *   The language to check for a path in or NULL.
    *
    * @return bool
-   *   TRUE if the path exists in that language.
+   *   TRUE if the path has an alias in the given language.
    */
-  public function doEvaluate($path, LanguageInterface $language = NULL) {
-    $language = !is_null($language) ? $language->getId() : NULL;
+  protected function doEvaluate($path, LanguageInterface $language = NULL) {
+    $language = is_null($language) ? NULL : $language->getId();
     $alias = $this->aliasManager->getAliasByPath($path, $language);
     return $alias != $path;
   }
--- a/src/Plugin/Condition/PathHasAlias.php
+++ b/src/Plugin/Condition/PathHasAlias.php
@@ -7,6 +7,7 @@
 
 namespace Drupal\rules\Plugin\Condition;
 
+use Drupal\Core\Language\LanguageInterface;
 use Drupal\Core\Path\AliasManagerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\rules\Core\RulesConditionBase;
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   * @param mixed $language
+   *   The language to check for a path in or NULL.
+   *
+   * @return bool
+   *   TRUE if the path exists in that language.
    */
-  public function evaluate() {
-    $path = $this->getContextValue('path');
-    $language = $this->getContext('language')->hasContextValue() ? $this->getContextValue('language')->getId() : NULL;
+  public function doEvaluate($path, LanguageInterface $language = NULL) {
```
klausi: should be protected, not public.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
...
+    $language = !is_null($language) ? $language->getId() : NULL;
```
klausi: this line is not necessary since the alias manager interface accepts NULL also.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Check if path has alias.
+   *
+   * @param string $path
+   *   The path to check.
+   * @param mixed $language
```
klausi: type hint should be \Drupal\Core\Language\LanguageInterface|null

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   *   The language to check for a path in or NULL.
+   *
+   * @return bool
+   *   TRUE if the path exists in that language.
```
klausi: should be "TRUE if the path has an alias in the given language."

src/Plugin/Condition/PathHasAlias.php
emclaughlin1215: What should it be, then? Because just taking it out creates an error because getAliasByPath expects a string, not a Language object. And putting $language->getId() in the getAliasByPath call errors out when $language is NULL, obviously.

klausi: ah, sorry, so then it is fine. Can you rename $language to $language_code then so that we don't overwrite the parameter which has a completely different type?

klausi: And we can avoid the "!" operator here if you just reverse the if/else branches.

emclaughlin1215:
Changes made per kalusi's comments.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   *   The language to check for a path in or NULL.
+   *
+   * @return bool
+   *   TRUE if the path has an alias in the given language.
    */
-  public function evaluate() {
-    $path = $this->getContextValue('path');
-    $language = $this->getContext('language')->hasContextValue() ? $this->getContextValue('language')->getId() : NULL;
+  protected function doEvaluate($path, LanguageInterface $language = NULL) {
+    $language = is_null($language) ? NULL : $language->getId();
```
klausi, line 87: Can we name this variable $langcode, same as on AliasManagerInterface? It is confusing to overwrite the parameter variable.

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Check if path has alias.
+   *
+   * @param string $path
+   *   The path to check.
+   * @param \Drupal\Core\Language\LanguageInterface|null $language
+   *   The language to check for a path in or NULL.
```
klausi, line 81: Should be "An optional language to look up the path in."

src/Plugin/Condition/PathHasAlias.php
```diff
@@ -72,11 +73,18 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Check if path has alias.
```
klausi, line 76: Should be "Check if a URL path has a URL alias."

klausi:
This has been fixed in the meantime

PR #264, opened 2015-10-09
Issue #2585419: Replace evaluate() with doEvaluate() for PathAliasExi…

emclaughlin1215:
…sts condition.

See d.o issues: https://www.drupal.org/node/2585419 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/PathAliasExists.php
+++ b/src/Plugin/Condition/PathAliasExists.php
@@ -78,15 +78,15 @@ class PathAliasExists extends RulesConditionBase implements ContainerFactoryPlug
    *
    * @param string $alias
    *   The alias to see if exists.
-   * @param mixed $language
+   * @param \Drupal\Core\Language\LanguageInterface|null $language
    *   The language to use.
    *
    * @return bool
    *   TRUE if the system path does not match the given alias (ie: the alias
    *   exists).
    */
-  public function doEvaluate($alias, LanguageInterface $language = NULL) {
-    $language = !is_null($language) ? $language->getId() : NULL;
+  protected function doEvaluate($alias, LanguageInterface $language = NULL) {
+    $language = is_null($language) ? NULL : $language->getId();
     $path = $this->aliasManager->getPathByAlias($alias, $language);
     return $path != $alias;
   }
--- a/src/Plugin/Condition/PathAliasExists.php
+++ b/src/Plugin/Condition/PathAliasExists.php
@@ -7,6 +7,7 @@
 
 namespace Drupal\rules\Plugin\Condition;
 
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/PathAliasExists.php
```diff
@@ -73,11 +74,19 @@ public static function create(ContainerInterface $container, array $configuratio
   }
 
   /**
-   * {@inheritdoc}
+   * Check if a path alias exists.
+   *
+   * @param string $alias
+   *   The alias to see if exists.
+   * @param mixed $language
```
klausi: type hint should be \Drupal\Core\Language\LanguageInterface|null

src/Plugin/Condition/PathAliasExists.php
```diff
@@ -73,11 +74,19 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   *
+   * @return bool
+   *   TRUE if the system path does not match the given alias (ie: the alias
+   *   exists).
    */
-  public function evaluate() {
-    $alias = $this->getContextValue('alias');
-    $language = $this->getContext('language')->hasContextValue() ? $this->getContextValue('language')->getId() : NULL;
+  public function doEvaluate($alias, LanguageInterface $language = NULL) {
+    $language = !is_null($language) ? $language->getId() : NULL;
```
klausi: no need for this line we can just pass on $language, the alias manager also accepts NULL.

emclaughlin1215: Just passing on $language creates the same problem mentioned in PR 263. Specifically, getPathByAlias() wants a string value for the language ID, not the language object itself, or NULL.

klausi: see comments in #263

emclaughlin1215:
Updated based on comments.

src/Plugin/Condition/PathAliasExists.php
```diff
@@ -73,11 +74,19 @@ public static function create(ContainerInterface $container, array $configuratio
...
+   *
+   * @return bool
+   *   TRUE if the system path does not match the given alias (ie: the alias
+   *   exists).
    */
-  public function evaluate() {
-    $alias = $this->getContextValue('alias');
-    $language = $this->getContext('language')->hasContextValue() ? $this->getContextValue('language')->getId() : NULL;
+  protected function doEvaluate($alias, LanguageInterface $language = NULL) {
+    $language = is_null($language) ? NULL : $language->getId();
```
klausi, line 89: Let's not overwrite the parameter variable $language here but rather use $lang_code instead.

klausi:
continued in #298

PR #265, opened 2015-10-09
Issue #2585429: Replacing evaluate() with doEvaluate() in NodeIsStick…

emclaughlin1215:
…y condition.

See d.o issues: https://www.drupal.org/node/2585429 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/NodeIsSticky.php
+++ b/src/Plugin/Condition/NodeIsSticky.php
@@ -36,7 +36,7 @@ class NodeIsSticky extends RulesConditionBase {
    * @return bool
    *   TRUE if the node is sticky.
    */
-  public function doEvaluate(NodeInterface $node) {
+  protected function doEvaluate(NodeInterface $node) {
     return $node->isSticky();
   }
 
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/NodeIsSticky.php
```diff
@@ -27,10 +28,15 @@
...
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node to check for stickiness.
+   *
+   * @return bool
+   *   TRUE if the node is sticky.
    */
-  public function evaluate() {
-    $node = $this->getContextValue('node');
+  public function doEvaluate(NodeInterface $node) {
```
klausi: should be protected, not public.

emclaughlin1215:

src/Plugin/Condition/NodeIsSticky.php
```diff
@@ -27,10 +28,15 @@
  */
 class NodeIsSticky extends RulesConditionBase {
   /**
-   * {@inheritdoc}
+   * Check if the given node is sticky.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node to check for stickiness.
```
klausi: "The node to check." should be enough.

emclaughlin1215:
Updated based on comments.

klausi:
merged, thanks!

PR #266, opened 2015-10-09
Issue #2585431: Replacing evaluate() with doEvaluate() in NodeIsPubli…

emclaughlin1215:
…shed condition.

See d.o issues: https://www.drupal.org/node/2585431 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/NodeIsPublished.php
+++ b/src/Plugin/Condition/NodeIsPublished.php
@@ -7,6 +7,7 @@
 
 namespace Drupal\rules\Plugin\Condition;
 
+use Drupal\node\NodeInterface;
 use Drupal\rules\Core\RulesConditionBase;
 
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/NodeIsPublished.php
```diff
@@ -28,10 +29,15 @@
 class NodeIsPublished extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Checks if a node is published.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node to check if is published.
```
klausi: "The node to check." should be enough.

emclaughlin1215:
Updated per comments.

klausi:
merged, thanks!

PR #267, opened 2015-10-09
Issue #2585435: Replacing evaluate() with doEvaluate() in NodeIsPromo…

emclaughlin1215:
…ted condition.

See d.o issues: https://www.drupal.org/node/2585435 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/NodeIsPromoted.php
+++ b/src/Plugin/Condition/NodeIsPromoted.php
@@ -7,6 +7,7 @@
 
 namespace Drupal\rules\Plugin\Condition;
 
+use Drupal\node\NodeInterface;
 use Drupal\rules\Core\RulesConditionBase;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/NodeIsPromoted.php
```diff
@@ -28,10 +29,15 @@
 class NodeIsPromoted extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Checks if a node is promoted.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node to check if is promoted.
```
klausi: "The node to check." should be enough.

emclaughlin1215:
Updated per comments.

klausi:
merged, thanks!

PR #268, opened 2015-10-09
Issue #2585463: Replacing evaluate() with doEvaluate() in NodeIsOfTyp…

emclaughlin1215:
…e condition.

See d.o issues https://www.drupal.org/node/2585463 and https://www.drupal.org/node/2572119

### DIFF
--- a/src/Plugin/Condition/NodeIsOfType.php
+++ b/src/Plugin/Condition/NodeIsOfType.php
@@ -42,7 +42,7 @@ class NodeIsOfType extends RulesConditionBase {
    * @return bool
    *   TRUE if the node type is in the array of types.
    */
-  public function doEvaluate(NodeInterface $node, array $types) {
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/NodeIsOfType.php
```diff
@@ -31,11 +32,17 @@
 class NodeIsOfType extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Check if a node is of a specific set of types.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node to check for a type.
+   * @param array $types
```
klausi: the data type should be "string[]", not just array.

emclaughlin1215:
Updated per comments.

klausi:
merged, thanks!

PR #270, opened 2015-10-10
Issue #2572119: Convert evaluate() to doEvaluate() for the EntityIsOf…

emclaughlin1215:
…Bundle condition.

### DIFF
--- a/src/Plugin/Condition/EntityIsOfBundle.php
+++ b/src/Plugin/Condition/EntityIsOfBundle.php
@@ -50,7 +50,7 @@ class EntityIsOfBundle extends RulesConditionBase {
    * @return bool
    *   TRUE if the provided entity is of the provided type and bundle.
    */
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/EntityIsOfBundle.php
```diff
@@ -37,12 +38,19 @@
...
+   *   The bundle to check for.
+   *
+   * @return bool
+   *   TRUE if the provided entity is of the provided type and bundle.
    */
-  public function evaluate() {
-    $provided_entity = $this->getContextValue('entity');
-    $specified_type = $this->getContextValue('type');
-    $specified_bundle = $this->getContextValue('bundle');
+  protected function doEvaluate(EntityInterface $provided_entity, $specified_type, $specified_bundle) {
```
klausi: Let's use $entity as name here to match the annotation, same for type and bundle.

emclaughlin1215:

klausi:
merged, thanks!

PR #271, opened 2015-10-10
Issue #2572119: Replace evaluate() with doEvaluate() in EntityIsNew c…

emclaughlin1215:
…ondition.

### DIFF
--- a/src/Plugin/Condition/EntityIsNew.php
+++ b/src/Plugin/Condition/EntityIsNew.php
@@ -38,7 +38,7 @@ class EntityIsNew extends RulesConditionBase {
    * @return bool
    *   TRUE if provided entity is new.
    */
-  public function doEvaluate(EntityInterface $provided_entity) {
+  protected function doEvaluate(EntityInterface $provided_entity) {
     return $provided_entity->isNew();
   }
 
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/EntityIsNew.php
```diff
@@ -29,10 +30,15 @@
...
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $provided_entity
+   *   The entity to check if is new.
+   *
+   * @return bool
+   *   TRUE if provided entity is new.
    */
-  public function evaluate() {
-    $provided_entity = $this->getContextValue('entity');
+  protected function doEvaluate(EntityInterface $provided_entity) {
```
klausi: Let's use $entity as name here to match the name in the annotation.

src/Plugin/Condition/EntityIsNew.php
```diff
@@ -29,10 +30,15 @@
 class EntityIsNew extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Check if the provided entity is new.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $provided_entity
+   *   The entity to check if is new.
```
klausi: "The entity to check." should be enough.

src/Plugin/Condition/EntityIsNew.php
```diff
@@ -29,10 +30,15 @@
...
+   *
+   * @return bool
+   *   TRUE if provided entity is new.
```
klausi: "... if the provided entity ..."

emclaughlin1215:

klausi:
merged, thanks!

PR #273, opened 2015-10-10
Issue #2572119: Replacing evaluate() with doEvaluate() in the DataLis…

emclaughlin1215:
…tCountIs condition.

### DIFF
--- a/src/Plugin/Condition/DataListCountIs.php
+++ b/src/Plugin/Condition/DataListCountIs.php
@@ -52,7 +52,7 @@ class DataListCountIs extends RulesConditionBase {
    * @return bool
    *   TRUE if the comparison returns true.
    */
-  public function doEvaluate($list, $operator, $value) {
+  protected function doEvaluate($list, $operator, $value) {
     switch ($operator) {
       case '==':
         return count($list) == $value;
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/DataListCountIs.php
```diff
@@ -40,13 +40,19 @@
...
+   *
+   * @param array $list
+   *   The list to compare the value to.
+   * @param string $operator
+   *   The type of comparison to do, maybe be one of '==', '<', or '>'.
+   * @param int $value
+   *   The value of that the count is to compare to.
+   *
+   * @return bool
+   *   TRUE if the comparison returns true.
```
klausi, line 53: EDIT: oops, since we have a comparison operator this is fine as is.

src/Plugin/Condition/DataListCountIs.php
```diff
@@ -40,13 +40,19 @@
 class DataListCountIs extends RulesConditionBase {
 
   /**
-   * {@inheritdoc}
+   * Compare the value to the count of the list.
+   *
+   * @param array $list
+   *   The list to compare the value to.
+   * @param string $operator
+   *   The type of comparison to do, maybe be one of '==', '<', or '>'.
```
klausi: I think you mean "may be" and not "maybe be"?

emclaughlin1215:

klausi:
merged, thanks!

PR #274, opened 2015-10-13
Issue #2591177: Rules definition objects are almost aways returning da…

torenware:
…ta type 'any'.

This is pretty preliminary, but I wanted this to be where folks can find it.

Drupal Issue https://www.drupal.org/node/2591177

### DIFF
--- a/src/Context/ContextDefinition.php
+++ b/src/Context/ContextDefinition.php
@@ -80,11 +80,11 @@ class ContextDefinition extends ContextDefinitionCore implements ContextDefiniti
     }
     // Default to Rules context definition class.
     $values['class'] = isset($values['class']) ? $values['class'] : '\Drupal\rules\Context\ContextDefinition';
-    if (!isset($values['type'])) {
-      $values['type'] = 'any';
+    if (!isset($values['value'])) {
+      $values['value'] = 'any';
     }
 
-    $definition = $values['class']::create($values['type']);
+    $definition = $values['class']::create($values['value']);
     foreach (array_intersect_key(static::$nameMap, $values) as $key => $name) {
       $definition->$name = $values[$key];
     }
--- a/tests/src/Integration/Engine/AnnotationProcessingTest.php
+++ b/tests/src/Integration/Engine/AnnotationProcessingTest.php
@@ -1,0 +1,126 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Tests\rules\Integration\Engine\AnnotationProcessingTest.
+ */
+
+namespace Drupal\Tests\rules\Integration\Engine;
+
+use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
+use Drupal\Core\Session\SessionManagerInterface;
+
+/**
+ * Tests processing of the ContextDefinition annotation.
+ *
+ * @group rules
+ */
+class AnnotationProcessingTest extends RulesIntegrationTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    $session_manager = $this->prophesize(SessionManagerInterface::class);
+    $this->container->set('session_manager', $session_manager->reveal());
+  }
+
+  /**
+   * Tests if our ContextDefinition annotations are correctly processed.
+   *
+   * @param string $plugin_type
+   *   Type of rules plugin to test (for now, 'action' or 'condition').
+   * @param string $plugin_id
+   *   Plugin ID for the plugin to be tested.
+   * @param string $context_name
+   *   The name of the plugin's context to test.
+   * @param string $expected
+   *   The type of context as defined in the plugin's annotation.
+   *
+   * @dataProvider provideRulesPlugins
+   *
+   * @group rules
+   */
+  public function testCheckConfiguration($plugin_type, $plugin_id, $context_name, $expected) {
+    $plugin = NULL;
+
+    switch ($plugin_type) {
+      case 'action':
+        $plugin = $this->actionManager->createInstance($plugin_id);
+        break;
+
+      case 'condition':
+        $plugin = $this->conditionManager->createInstance($plugin_id);
+        break;
+    }
+
+    $this->assertNotNull($plugin, "{$plugin_type} plugin {$plugin_id} loads");
+
+    $context = $plugin->getContext($context_name);
+
+    $this->assertNotNull($context, "Plugin {$plugin_id} has context {$context_name}");
+
+    $context_def = $context->getContextDefinition();
+    $type = $context_def->getDataType();
+
+    $this->assertSame($type, $expected, "Context type for {$context_name} is $expected");
+  }
+
+  /**
+   * Provider for plugins to test.
+   *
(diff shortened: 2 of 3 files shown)


torenware:
This one is no-go; looks like I broke the deriver.  When I figure this out, I'll push an update.

torenware:
It turns out my fix was fine.  The reasons the tests broke is that until now, our mocks have only handled ContextDefinitions with a DataType of 'any', so our entity type mocks needed to be expanded to cover the types of entity:user, entity:user_role, and entity:node. This is in addition to the entity:test type that we were already mocking.

torenware:
Per @klausi, I've added an integration test to test whether Rules plugins correctly report their data type.

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -0,0 +1,136 @@
...
+
+      case 'condition':
+        $plugin = $this->conditionManager->createInstance($params['plugin']);
+        break;
+    }
+
+    $this->assertNotNull($plugin, "{$params['type']} plugin {$params['plugin']} loads");
+    if (empty($plugin)) {
+      return;
+    }
```
klausi: why is that if() necessary? I think phpunit already stops when an assertion fails.

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -0,0 +1,136 @@
...
+   * Tests if our ContextDefiniton annotations are correctly processed.
+   *
+   * @dataProvider provideRulesPlugins
+   *
+   * @param array $params
+   *   An associative array containing type information from Rules plugins.
+   *
+   * @group rules
+   */
+  public function testCheckConfiguration($params) {
```
klausi: instead of $params can we separate out all 4 parameters to make the test easier to read?

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -0,0 +1,136 @@
...
+    if (empty($plugin)) {
+      return;
+    }
+
+    $context = $plugin->getContext($params['context']);
+
+    $this->assertNotNull($context, "Plugin {$params['plugin']} has context {$params['context']}");
+    if (empty($context)) {
+      return;
+    }
```
klausi: another if() that can be removed.

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -0,0 +1,136 @@
...
+ * Contains \Drupal\Tests\rules\Integration\Engine\AnnotationProcessingTest.php.
+ */
+
+namespace Drupal\Tests\rules\Integration\Engine;
+
+use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
+use Drupal\Core\Session\SessionManagerInterface;
+
+/**
+ * Test auto saving of variables after Rules execution.
```
klausi: Comment is wrong

tests/src/Integration/RulesEntityIntegrationTestBase.php
```diff
@@ -53,15 +53,47 @@ public function setUp() {
...
+        'entity_keys' => [
+          'bundle' => 'dummy',
+        ],
+      ],
+    ];
+
+    $type_array = [];
+
+    foreach ($type_info as $type => $info) {
+      $entity_type = new ContentEntityType($info);
```
klausi: a user role is not a content entity type, so this does not really fit.

src/Context/ContextDefinition.php
```diff
@@ -81,7 +81,12 @@ public static function createFromArray($values) {
     // Default to Rules context definition class.
     $values['class'] = isset($values['class']) ? $values['class'] : '\Drupal\rules\Context\ContextDefinition';
     if (!isset($values['type'])) {
-      $values['type'] = 'any';
+      if (!isset($values['value'])) {
+        $values['type'] = 'any';
+      }
+      else {
+        $values['type'] = $values['value'];
+      }
```
klausi: why do we have to do this? Can't we just change $nameMap and point 'value' => 'dataType'? Or are we using 'type' as key anywhere else?

torenware: @klausi -- yeah, we are using values['type'] down stream. I'm can find you the references if you want.  Changing them from 'type' to 'value' should probably be a separate issue, however.

tests/src/Integration/RulesEntityIntegrationTestBase.php
torenware: We do need a mock for it, so I'll make sure that it's passed as a ConfigEntityType.

tests/src/Integration/Engine/AnnotationProcessingTest.php
torenware: Fixed.

klausi: forgot to push?

torenware: Nope, I pushed it.  If you view the file with the View button, you'll see I made the fix.  For some reason, though, the diff shows the old state of the file.  Github bug I'm guessing.

torenware:
OK, looks like I fixed the wrong comment.  Fixed both now.

tests/src/Integration/RulesEntityIntegrationTestBase.php
```diff
@@ -53,15 +54,48 @@ public function setUp() {
...
+      $type_array[$type] = $entity_type;
+    }
+
+    // We need a user_role mock as well.
+    $role_entity_info = [
+      'id' => 'user_role',
+      'label' => 'Test Role',
+    ];
+    $role_type = new ConfigEntityType($role_entity_info);
+    $type_array['user_role'] = $role_type;
```
torenware: Note that the role is now handled as a ConfigEntityType.

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -0,0 +1,126 @@
...
+ */
+class AnnotationProcessingTest extends RulesIntegrationTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    $this->sessionManager = $this->prophesize(SessionManagerInterface::class);
```
klausi: all class variables should be declared and documented on the class. And why would an annotation test need a session manager? Please add a comment.

torenware: The issue here is that I'm loading a set of plugins and testing the annotations as used in those plugins.  If a plugin has an entity:user context, the plugin will invoke APIs that will require a Session Manager mock.  

When constructing the mocks, I was letting Prophecy tell me what methods of what interfaces it required.  It apparently needed SessionMangerInterface available in services container.

I'll check why $this->sessionManager is not declared here, since I'd agree that it needs to be there if I refer to it.  It may not be necessary, however, since I suspect that adding the service to the container will suffice.

klausi:
merged, thanks!

PR #275, opened 2015-10-15
Issue #2592891: Updating the execute() and evaluate() functions in Ru…

emclaughlin1215:
…lesActionBase and RulesConditionBase to not add NULL values for contexts that do not exist; fixing a test that did not provide a default for a non-required value.

See d.o issue https://www.drupal.org/node/2593891

Also: execute() and evaluate() are using different methods to get the context values for the array. Is this right? Or should I update them to be the same. And if I should update them, which one should I update? I'm not sure which one is more efficient or what.


klausi:
See https://www.drupal.org/node/2593891#comment-10510280

PR #281, opened 2015-11-06
Isssue #2609816: @group annotation for ALL THE TESTS.

torenware:
Drupal issue https://www.drupal.org/node/2609816 (Rules tests should be "testbot friendly").

### DIFF
--- a/tests/src/Integration/Condition/ConditionManagerTest.php
+++ b/tests/src/Integration/Condition/ConditionManagerTest.php
@@ -11,6 +11,8 @@ use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
 
 /**
  * Tests the Rules condition manager.
+ *
+ * @group rules_conditions
  */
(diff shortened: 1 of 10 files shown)


tests/src/Integration/Condition/ConditionManagerTest.php
```diff
@@ -11,6 +11,8 @@
 
 /**
  * Tests the Rules condition manager.
+ * @group rules_conditions
```
martinfrances107: Overall I think this pull request is a good idea.

The initiative to make rules - "testbot friendly" is a good idea.

This is a minor nit pick there should be a blank line between the text and the group tag in the code fragment above.

It will show up on a phpcs error report.

torenware: Yep, looks like I "missed a spot".  Will fix.

klausi:
merged, thanks!

PR #282, opened 2015-11-06
Issue #2610458: Squelch annotation objects.

torenware:
Drupal Issue https://www.drupal.org/node/2610458.

<del>I have a test for this, but it makes sense to integrate it into the test class created for #2591177
(Rules context processing coerces most context types to 'any').</del>

Includes a test.

### DIFF
--- a/src/Context/Annotation/ContextDefinition.php
+++ b/src/Context/Annotation/ContextDefinition.php
@@ -8,6 +8,7 @@
 namespace Drupal\rules\Context\Annotation;
 
 use Drupal\Core\Annotation\ContextDefinition as CoreContextDefinition;
+use Drupal\Core\Annotation\Translation;
 use Drupal\rules\Context\ContextDefinition as RulesContextDefinition;
 
 /**
@@ -33,6 +34,12 @@ class ContextDefinition extends CoreContextDefinition {
    * {@inheritdoc}
    */
   public function __construct(array $values) {
+    // Filter out any @Translation annotation objects.
+    foreach ($values as $key => $value) {
+      if ($value instanceof Translation) {
+        $values[$key] = $value->get();
+      }
+    }
     $this->definition = RulesContextDefinition::createFromArray($values);
   }
 
@@ -40,6 +47,7 @@ class ContextDefinition extends CoreContextDefinition {
    * Returns the value of an annotation.
    *
    * @return \Drupal\rules\Context\ContextDefinitionInterface.
+   *   Return the Rules version of the ContextDefinitionInterface.
    */
   public function get() {
     return $this->definition;
--- a/tests/src/Integration/Engine/AnnotationProcessingTest.php
+++ b/tests/src/Integration/Engine/AnnotationProcessingTest.php
@@ -7,8 +7,9 @@
 
 namespace Drupal\Tests\rules\Integration\Engine;
 
-use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
+use Drupal\Core\Annotation\Translation;
 use Drupal\Core\Session\SessionManagerInterface;
+use Drupal\Tests\rules\Integration\RulesIntegrationTestBase;
 
 /**
  * Tests processing of the ContextDefinition annotation.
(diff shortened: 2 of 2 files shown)


tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -23,11 +23,46 @@ class AnnotationProcessingTest extends RulesIntegrationTestBase {
...
+    $context = $plugin->getContext('list');
+    $definition = $context->getContextDefinition();
+
+    // These can reasonable be string or TranslatableMarkup, but never
+    // a Translation object.
+    $label = $definition->getLabel();
+    $description = $definition->getDescription();
+
+    $valid = !(is_object($label) && ($label instanceof Translation));
+    $this->assertTrue($valid, 'Label is not a Translation object');
```
klausi: can we just use is_string() instead?

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -23,11 +23,46 @@ class AnnotationProcessingTest extends RulesIntegrationTestBase {
...
+
+    $valid = !(is_object($description) && ($description instanceof Translation));
+    $this->assertTrue($valid, 'Description is not a Translation object');
```
klausi: same here, should be is_string()

tests/src/Integration/Engine/AnnotationProcessingTest.php
torenware: @klausi  Let me think... it's valid if it's a string, or if it's TranslableMarkup.  So actually, the test is "it's not a Translation object".  I'm probably testing is_object($label) on the assumption that the instanceof operator won't be happy if I pass it a non-object. Which might not be a valid assumption, BTW.

So, no, we can't use is_string().  But there may be a better test than what I'm using here, nonetheless.

I just did a test: `"string" instanceof CLASS` returns FALSE, and does not cause an error.  So I can reasonably simplify the test to

```
$valid = !($label instanceof Translation)
```

That work better for you?

tests/src/Integration/Engine/AnnotationProcessingTest.php
torenware: Same as before -- can't use is_string(), because TranslatableMarkup is perfectly reasonable, and may be what we get under some circumstances.

tests/src/Integration/Engine/AnnotationProcessingTest.php
```diff
@@ -23,11 +23,38 @@ class AnnotationProcessingTest extends RulesIntegrationTestBase {
...
+    // Get a sample Rules plugin.
+    $plugin = $this->conditionManager->createInstance('rules_list_contains');
+    $context = $plugin->getContext('list');
+    $definition = $context->getContextDefinition();
+
+    // These can reasonable be string or TranslatableMarkup, but never
+    // a Translation object.
+    $label = $definition->getLabel();
+    $description = $definition->getDescription();
+    $this->assertFalse($label instanceof Translation, 'Label is not a Translation object');
```
klausi, line 45: You forgot to include a use statement for the Translation class, so this will always be false.

src/Context/Annotation/ContextDefinition.php
```diff
@@ -9,6 +9,7 @@
 
 use Drupal\Core\Annotation\ContextDefinition as CoreContextDefinition;
 use Drupal\rules\Context\ContextDefinition as RulesContextDefinition;
+use Drupal\Core\Annotation\Translation;
```
klausi, line 11: let's use alphabetical ordering for use statements.

src/Context/Annotation/ContextDefinition.php
```diff
@@ -33,6 +34,13 @@ class ContextDefinition extends CoreContextDefinition {
    * {@inheritdoc}
    */
   public function __construct(array $values) {
+    // Filter out any @Translation annotation objects.
+    $old_values = $values;
```
klausi: You are overwriting $values anyway, so the $old_values variable is not needed?

torenware: I'm a little reluctant to write into a PHP array that's used in a foreach() loop, lest I cause something to reset :-)  If that's safe, then I can remove that.

tests/src/Integration/Engine/AnnotationProcessingTest.php
torenware, line 45: True that.  Will fix.

klausi:
merged, thanks!

PR #283, opened 2015-11-07
Issue #2610574: Nested expressions throw exceptions when saved.

torenware:
Drupal issue https://www.drupal.org/node/2610574.

Fix with test.

### DIFF
--- a/tests/src/Kernel/ConfigEntityTest.php
+++ b/tests/src/Kernel/ConfigEntityTest.php
@@ -8,6 +8,7 @@
 namespace Drupal\Tests\rules\Kernel;
 
 use Drupal\rules\Context\ContextDefinition;
+use Drupal\rules\Context\ContextConfig;
 
 /**
  * Tests storage and loading of Rules config entities.
@@ -22,6 +23,15 @@ class ConfigEntityTest extends RulesDrupalTestBase {
    * @var \Drupal\Core\Entity\EntityStorageInterface
    */
   protected $storage;
+
+  /**
+   * Disable strict config schema checking for now.
+   *
+   * @todo: Fix once config schema has been improved.
+   *
+   * @var bool
+   */
+  protected $strictConfigSchema = FALSE;
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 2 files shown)


torenware:
<del>The Travis fail is expected: the test reveals the bug, but the PR does not yet have a fix attached.</del>

<em>As of 8 November, the fix is in.</em>

tests/src/Kernel/ConfigEntityTest.php
```diff
@@ -66,6 +67,41 @@ public function testConfigAction() {
...
+    $rule->addAction('rules_test_log');
+
+    // Save it as a config_entity.
+    $rule_config = $this->storage->create(['id' => 'test_big_hairy_rule']);
+    $rule_config->setExpression($rule);
+    $rule_config->set('event', 'rules_user_login');
+    $rule_config->set('tag', 'demo example');
+    $rule_config->set('description', 'Reaction Rule That tests if context gets passed from event');
+    $rule_config->save();
+
```
torenware, line 112: Currently, we throw an exception at line 97.

tests/src/Kernel/ConfigEntityTest.php
```diff
@@ -66,6 +76,46 @@ public function testConfigAction() {
...
+      ->addCondition('rules_test_false')
+      ->addCondition('rules_test_true', ContextConfig::create()->negateResult())
+      ->negate();
+
+    $or = $this->expressionManager->createOr()
+      ->addCondition('rules_test_true', ContextConfig::create()->negateResult())
+      ->addCondition('rules_test_false')
+      ->addCondition($and);
+
+    $rule = $this->expressionManager->createRule();
```
fago: createReactionRule()  - the plugin does not fit the config entity else ;) I guess we should check the class somewhere in the config enttiy impl?

torenware: We can only change our own config entity implementation, of course.  We need to remove anything other than arrays and primitives before we call the parent:: methods, or we get the original crash.

I can certainly move the code out of the expression hierarchy into the config entity class, but IIRC it's a bit easier if I can see protected data; the methods are simpler and easier to understand.  I'll take another look at this to see if this is

tests/src/Kernel/ConfigEntityTest.php
```diff
@@ -66,6 +76,46 @@ public function testConfigAction() {
...
+    // Create a nested expression.
+    $and = $this->expressionManager->createAnd()
+      ->addCondition('rules_test_false')
+      ->addCondition('rules_test_true', ContextConfig::create()->negateResult())
+      ->negate();
+
+    $or = $this->expressionManager->createOr()
+      ->addCondition('rules_test_true', ContextConfig::create()->negateResult())
+      ->addCondition('rules_test_false')
+      ->addCondition($and);
```
klausi, line 96: This is wrong, the addCondition() method takes a string parameter (the condition id). You want addExpressionObject() instead.

torenware, line 96: Changing this does make the test pass w/o the fix code.

So it was an error, but a subtle error.  We probably should enforce is_string() on createAction() and createCondition(), or at least use type hinting to get PHP to put out a warning.  This way, the problem is revealed at the source, rather than only when the config entity is saved.

torenware, line 96: Scalar type hinting is still not supported in PHP 5.x (I'm not sure about PHP 7), but the following would catch this kind of error in a way that's much easier to debug:

```
  /**
   * {@inheritdoc}
   */
  public function createCondition($id) {
    if (!is_string($id)) {
      throw new \InvalidArgumentException;
    }
    return $this->createInstance('rules_condition', [
      'condition_id' => $id,
    ]);
  }
```

This throws an exception at the point of the error:

```
There was 1 error:

1) Drupal\Tests\rules\Kernel\ConfigEntityTest::testNestedRuleSave
InvalidArgumentException:

/var/www/drupal/modules/custom/rules/src/Engine/ExpressionManager.php:66
/var/www/drupal/modules/custom/rules/src/Engine/ConditionExpressionContainer.php:88
/var/www/drupal/modules/custom/rules/tests/src/Kernel/ConfigEntityTest.php:96
```

What do you think about doing this to prevent this kind of thing?  The code in the test is based on stuff that was in the tree back in September, so these are easy kinds of errors to make. Doing this would make these kinds of errors easier to debug.

klausi, line 96: Yes , I think adding such a check + a meaningful message for the InvalidArgumentException would be a good improvement.

klausi:
Closing, see https://www.drupal.org/node/2610574#comment-10620322

PR #285, opened 2015-11-07
I2409053

piyuesh23:
See https://www.drupal.org/node/2409053

### DIFF
--- a/rules.module
+++ b/rules.module
@@ -9,6 +9,7 @@ use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Url;
 use Drupal\rules\Event\EntityEvent;
 use Drupal\rules\Event\UserLoginEvent;
 use Drupal\rules\Event\UserLogoutEvent;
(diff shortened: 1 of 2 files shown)


src/Plugin/RulesAction/DataConvert.php
```diff
@@ -104,4 +104,13 @@ public function execute() {
     $this->setProvidedValue('conversion_result', $result);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function refineContextDefinitions() {
+    if ($type = $this->getContextValue('target_type')) {
+      $this->pluginDefinition['provides']['conversion_result']->setDataType($type);
+    }
```
torenware, line 113: A before/after test should be added to show that refining the type actually creates a change in type.

rules.module
```diff
@@ -20,8 +21,8 @@ function rules_help($route_name, RouteMatchInterface $route_match) {
   switch ($route_name) {
     case 'entity.rules_reaction_rule.collection':
       $output = t('Reaction rules, listed below, react on selected events on the site. Each reaction rule may fire any number of <em>actions</em>, and may have any number of <em>conditions</em> that must be met for the actions to be executed. You can also set up <a href="!url1">components</a> – stand-alone sets of Rules configuration that can be used in Rules and other parts of your site. See <a href="!url2">the online documentation</a> for an introduction on how to use Rules.', [
-        '!url1' => \Drupal::url('entity.rules_component.collection'),
-        '!url2' => 'http://drupal.org/node/298480',
+        '!url1' => Url::fromRoute('entity.rules_component.collection'),
+        '!url2' => Url::fromUri('http://drupal.org/node/298480'),
```
torenware, line 25: The '!' prefix is no longer valid since about 28 September; see https://www.drupal.org/node/2575819.  You should use ':url1' and ':url2' instead.  I'm also not sure if you should be using Url::fromUri() for ':url2', but if it works, it works.  Note that the Url class can do funny things.  Last -- Drupal URLs are https://.

dasjo:
here's the issue link: https://www.drupal.org/node/2409053

klausi:
continued in #325

PR #286, opened 2015-11-09
Issure #2611942: Adding a default value for all Conditions and Action…

emclaughlin1215:
…s that have context values where required is set to FALSE.

Not sure if I should've added default_value for ones where the default is NULL, but I figured adding it was better for consistency than not adding it.


klausi:
continued in #303

PR #288, opened 2015-11-25
Issue #2587179 by claudiu.cristea, Torenware: Place block operation i…

claudiu-cristea:
…n the modal Place block dialog does not bring up config dialog

See https://www.drupal.org/node/2587179

### DIFF
--- a/src/Context/ContextDefinition.php
+++ b/src/Context/ContextDefinition.php
@@ -7,7 +7,8 @@
 
 namespace Drupal\rules\Context;
 
-use \Drupal\Core\Plugin\Context\ContextDefinition as ContextDefinitionCore;
+use Drupal\Core\Plugin\Context\ContextDefinition as ContextDefinitionCore;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
 
 /**
  * Extends the core context definition class with useful methods.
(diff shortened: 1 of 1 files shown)


torenware:
@claudiu-cristea we may want to compare / merge this one with Drupal Issue https://www.drupal.org/node/2610458, PR #282.  Your approach tests for TranslatableMarkup, which actually is not the problem; the thing we need to stop is the @Translation annotation.

src/Context/ContextDefinition.php
```diff
@@ -78,6 +79,21 @@ public static function createFromArray($values) {
...
       throw new \Exception('ContextDefinition class must implement \Drupal\rules\Context\ContextDefinitionInterface.');
     }
+
+    // Annotation classes extract data from passed annotation classes directly
+    // used in the classes they pass to.
+    foreach (['label', 'description'] as $key) {
+      // @todo Remove this workaround in https://www.drupal.org/node/2362727.
+      if (isset($values[$key])) {
+        if ($values[$key] instanceof TranslatableMarkup) {
+          $values[$key] = (string) $values[$key];
```
torenware, line 89: @claudiu-cristea we should instead test for `instanceof \Drupal\Core\Annotation\Translation`.  See https://www.drupal.org/node/2610458 for a test that demostrates the problem as well.

klausi:
see https://www.drupal.org/node/2587179#comment-10617522

PR #292, opened 2015-11-25
Remove usage of deprecated !placeholder

stefanruijsenaars:
!placeholder is deprecated and using it in hook_help generates a user error:
before

We can use :placeholder instead for URLs. In manual testing this patch gets rid of the error and the HTML is still correct:

after


klausi:
merged, thanks!

PR #293, opened 2015-11-25
Move to separately execute RulesComponent.

fago:


src/Engine/RulesStateInterface.php
```diff
@@ -22,6 +22,8 @@
    *   The variable name.
    * @param \Drupal\Core\TypedData\TypedDataInterface $data
    *   The variable wrapped as typed data.
+   *
+   * @return static
```
klausi: should be return $this, right?

src/Engine/RulesComponent.php
```diff
@@ -0,0 +1,151 @@
...
+
+  /**
+   * Constructs the object.
+   *
+   * @param \Drupal\rules\Engine\ExpressionInterface $expression
+   *   The expression of the component.
+   */
+  protected function __construct($expression) {
+    $this->state = RulesState::create();
+    $this->typedDataManager = \Drupal::typedDataManager();
```
klausi: meh, no dependency injection. Can we use a method getTypedDataManager() that calls out globally if the typedDataManager class variable is not set? Then we can also have a setTypedDataManager() method that allows setter injection. Better than nothing.

src/Engine/RulesComponent.php
```diff
@@ -0,0 +1,151 @@
...
+    $this->expression = $expression;
+  }
+
+  /**
+   * Adds a context definition.
+   *
+   * @param $name
```
klausi: data type missing, probably string?

src/Engine/RulesComponent.php
```diff
@@ -0,0 +1,151 @@
...
+    return new static($expression);
+  }
+
+  /**
+   * Constructs the object.
+   *
+   * @param \Drupal\rules\Engine\ExpressionInterface $expression
+   *   The expression of the component.
+   */
+  protected function __construct($expression) {
```
klausi: type hint for $expression missing

fago:
This is mostly done, only one @ todo for provided Context definitions needs to be addressed. Also, phpcs is wrong about @ return $this coding style, so looks like we have to ignore this rule also. Finally there are some risky tests here, not sure why I don't see them locally.

klausi:
A risky test is one that does not have an assertion. I think this was recently introduced to the phpunit config drupal 8 core.

src/Engine/RulesComponent.php
```diff
@@ -0,0 +1,169 @@
...
+   */
+  protected $expression;
+
+  /**
+   * Constructs the object.
+   *
+   * @param \Drupal\rules\Engine\ExpressionInterface $expression
+   *   The expression of the component.
+   *
+   * @return static
```
klausi, line 51: should be "@return $this", that's why phpcs is failing.

src/Engine/RulesComponent.php
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+namespace Drupal\rules\Engine;
+
+use Drupal\rules\Context\ContextDefinitionInterface;
+
+/**
+ * Handles executable Rules components.
+ */
+class RulesComponent {
```
klausi, line 15: we also have the RulesComponent.php config entity, which might cause confusion.

src/Engine/RulesState.php
```diff
@@ -163,6 +217,7 @@ public function autoSave() {
         $typed_data->getRoot()->getValue()->save();
       }
     }
+    return $this;
   }
```
klausi: The method setTypedDataManager() is missing to complete the setter injection pattern here when somebody wants to use/test this with a different typed data manager.

src/Engine/RulesStateInterface.php
```diff
@@ -75,11 +107,15 @@ public function applyDataSelector($selector, $langcode = LanguageInterface::LANG
    * @param string $selector
    *   The data selector that specifies the target object to be saved. Example:
    *   node:uid:entity.
+   *
+   * @return static
```
klausi: $this

src/Engine/RulesStateInterface.php
```diff
@@ -75,11 +107,15 @@ public function applyDataSelector($selector, $langcode = LanguageInterface::LANG
...
    */
   public function saveChangesLater($selector);
 
   /**
    * Saves all variables that have been marked for auto saving.
+   *
+   * @return static
```
klausi: $this

src/Entity/RulesComponent.php
```diff
@@ -96,6 +98,13 @@ class RulesComponent extends ConfigEntityBase {
   protected $expression_id;
 
   /**
+   * Array of context definitions arrays, keyed by context name.
```
klausi: should be "context definition arrays"

src/Entity/RulesComponent.php
```diff
@@ -147,6 +156,64 @@ public function setExpression(ExpressionInterface $expression) {
   }
 
   /**
+   * Gets the configured component.
+   *
+   * @return \Drupal\rules\Engine\RulesComponent
+   *   The component.
+   */
+  public function getComponent() {
+    $component = \Drupal\rules\Engine\RulesComponent::create($this->getExpression());
```
klausi, line 165: class name should be converted to use statement.

src/Entity/RulesComponent.php
```diff
@@ -147,6 +156,64 @@ public function setExpression(ExpressionInterface $expression) {
...
+
+  /**
+   * Sets the Rules component to be stored.
+   *
+   * @param \Drupal\rules\Engine\RulesComponent $component
+   *   The component.
+   *
+   * @return $this
+   */
+  public function setComponent(\Drupal\rules\Engine\RulesComponent $component) {
```
klausi, line 180: use statement

src/EventSubscriber/GenericEventSubscriber.php
```diff
@@ -78,21 +91,28 @@ public function onRulesEvent(Event $event, $event_name) {
...
+        );
       }
+    }
 
-      $reaction_rule->execute();
+    // Loop over all rules and execute them.
+    foreach ($configs as $config) {
+      /** @var ReactionRule $config */
+      $config->getExpression()
+        ->executeWithState($state);
```
klausi, line 112: Hm, does that mean that if a rule uses the add a variable action that the variable remains in the state and bleeds over to next rule that is executed?

fago, line 112: Yes. that's indeed an issue, We probably should solve that with clones as in d7 and add some test coverage for that

src/Engine/RulesComponent.php
fago, line 51: I looked it up - this is correct although our docs are not 100% accurate. opened https://www.drupal.org/node/2631930

src/Engine/RulesComponent.php
fago, line 15: Yes. We need to rename this to RulesComponentConfig. As discussed, let's do this in a follow-up.

src/Entity/RulesComponent.php
fago, line 165: not easily possible as it's the name of the current class - let's leave this to the re-name follow-up

src/Entity/RulesComponent.php
fago, line 180: see above

src/EventSubscriber/GenericEventSubscriber.php
fago, line 112: Tried fixing it, but turns out we miss that functinoality still in d8. Thus, I created an issue instead: https://www.drupal.org/node/2631948 We can work on this as part of the loop or component execution derivative support tasks.

PR #296, opened 2015-11-28
Feature/2624846 event system cron finished

ndf:
Hi Fago,

This is the Event for System : Cron maintenance tasks are performed

First I named it SystemCronEventFinished, but it is actually called before processing the cron-job.
hook_cron() does not give us any parameters to pass to the GenericEvent constructor.
There is 1 kernel-test, that call the rules_cron(). This is based on user_login event.

https://www.drupal.org/node/2624846


tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -101,4 +101,32 @@ public function testUserLogoutEvent() {
...
+    ]);
+    $config_entity->save();
+
+    // Rebuild the container so that the newly configured event gets picked up.
+    $this->container->get('kernel')->rebuildContainer();
+    // The logger instance has changed, refresh it.
+    $this->logger = $this->container->get('logger.channel.rules');
+
+    // Invoke the hook manually which should trigger the rule.
+    rules_cron();
```
ndf: talked with klausi: if available could load Cron service and then run it.
aka invoke a real cron-job

$this->container->get('cron')->run();

klausi:
merged, thanks!

PR #299, opened 2015-11-29
Updating DataListItemAdd action to use doExecute().

karolkasas:
Updated the execute() function to doExecute() for the DataListItemAdd class.
Issue #2472815

### DIFF
--- a/src/Plugin/RulesAction/DataListItemAdd.php
+++ b/src/Plugin/RulesAction/DataListItemAdd.php
@@ -58,7 +58,6 @@ class DataListItemAdd extends RulesActionBase {
    */
   protected function doExecute($list, $item, $unique, $pos) {
     $position = ($pos ? $pos : 'end');
-    $unique = ($unique ? $unique : FALSE);
     // Optionally, only add the list item if it is not yet contained.
     if (!((bool) $unique && in_array($item, $list))) {
       if ($position === 'start') {
--- a/src/Plugin/RulesAction/DataListItemAdd.php
+++ b/src/Plugin/RulesAction/DataListItemAdd.php
@@ -44,13 +44,21 @@ use Drupal\rules\Core\RulesActionBase;
 class DataListItemAdd extends RulesActionBase {
 
   /**
-   * {@inheritdoc}
+   * Executes the Plugin.
+   *
+   * @param array $list
+   *   An array to add an item to.
+   * @param mixed $item
+   *   An item which is being added to the array.
+   * @param bool $unique
+   *   A boolean value which indicates if the item should be updated if it is
+   *   already in the list array.
+   * @param string $pos
+   *   Position in array where the item is being added.
    */
-  public function execute() {
-    $list = $this->getContextValue('list');
-    $item = $this->getContextValue('item');
-    $position = ($this->getContextValue('pos') ? $this->getContextValue('pos') : 'end');
-    $unique = ($this->getContextValue('unique') ? $this->getContextValue('unique') : FALSE);
+  public function doExecute($list, $item, $unique, $pos) {
+    $position = ($pos ? $pos : 'end');
+    $unique = ($unique ? $unique : FALSE);
     // Optionally, only add the list item if it is not yet contained.
     if (!((bool) $unique && in_array($item, $list))) {
       if ($position === 'start') {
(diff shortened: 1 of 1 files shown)


src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,13 +44,21 @@
...
+   *   already in the list array.
+   * @param string $pos
+   *   Position in array where the item is being added.
    */
-  public function execute() {
-    $list = $this->getContextValue('list');
-    $item = $this->getContextValue('item');
-    $position = ($this->getContextValue('pos') ? $this->getContextValue('pos') : 'end');
-    $unique = ($this->getContextValue('unique') ? $this->getContextValue('unique') : FALSE);
+  public function doExecute($list, $item, $unique, $pos) {
```
klausi: this should be protected

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,13 +44,21 @@
...
+    $position = ($pos ? $pos : 'end');
+    $unique = ($unique ? $unique : FALSE);
```
klausi: I think we can just use $unique = (bool) $unique;

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,13 +44,21 @@
...
     // Optionally, only add the list item if it is not yet contained.
     if (!((bool) $unique && in_array($item, $list))) {
```
klausi, line 64: Then we don't need the (bool) cast here

karolkasas:
The test are passing without that line of code just fine, I have also fixed the protected access for the function. @klausi

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,13 +44,20 @@
...
+   *
+   * @param array $list
+   *   An array to add an item to.
+   * @param mixed $item
+   *   An item which is being added to the array.
+   * @param bool $unique
+   *   A boolean value which indicates if the item should be updated if it is
+   *   already in the list array.
+   * @param string $pos
+   *   Position in array where the item is being added.
```
fago: This should document the available options as well that it it's optional. Also is unique not optional also?

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,13 +44,20 @@
...
    */
-  public function execute() {
-    $list = $this->getContextValue('list');
-    $item = $this->getContextValue('item');
-    $position = ($this->getContextValue('pos') ? $this->getContextValue('pos') : 'end');
-    $unique = ($this->getContextValue('unique') ? $this->getContextValue('unique') : FALSE);
+  protected function doExecute($list, $item, $unique, $pos) {
```
fago: should repeat the default values here

karolkasas: Eh you are right, I forgot about the default values. Will fix this when I will be behind com.

src/Plugin/RulesAction/DataListItemAdd.php
karolkasas: Sorry I can't follow, you mean that the param comments should contain that the last two params are optional?

ndeet: I think he means that in the description of the param you should prefix the optional params with e.g. `(optional) Position in array ...`

karolkasas:
I have finally found time and fixed the optional params and comments for the params.

karolkasas:
@klausi @fago does this need additional work still?

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,16 +44,25 @@
...
+   *   An array to add an item to.
+   * @param mixed $item
+   *   An item which is being added to the array.
+   * @param bool $unique
+   *   (optional) A boolean value which indicates if the item should be updated
+   *   if it is already in the list array.
+   * @param string $pos
+   *   (optional) Position in array where the item is being added. Accepted
+   *   value are:
+   *   - "start"      Beginning of the array.
```
klausi, line 59: can we use a colon after the value example similar to https://www.drupal.org/coding-standards/docs#lists and without the additional spacing?

src/Plugin/RulesAction/DataListItemAdd.php
```diff
@@ -44,16 +44,25 @@
...
+   *   - "end"        End of the array.
```
klausi, line 60: same here

klausi:
continued in #326

PR #301, opened 2015-11-29
Updating DataSet action to use doExecute().

karolkasas:
Updated the execute() function to doExecute() for the DataSet class.
Issue #2472815


fago:
thx, merged it.

PR #304, opened 2015-11-29
2572119 data convert action

ndeet:

### DIFF
--- a/src/Plugin/RulesAction/DataConvert.php
+++ b/src/Plugin/RulesAction/DataConvert.php
@@ -49,9 +49,9 @@ class DataConvert extends RulesActionBase {
    * @param string $target_type
    *   The target for the action.
    * @param string $rounding_behavior
-   *   The behaviour for rounding.
+   *   The target type the value should be converted into.
    */
-  public function doExecute($value, $target_type, $rounding_behavior) {
+  protected function doExecute($value, $target_type, $rounding_behavior) {
     // @todo: Add support for objects implementing __toString().
     if (!is_scalar($value)) {
       throw new \InvalidArgumentException('Only scalar values are supported.');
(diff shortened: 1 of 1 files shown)


src/Plugin/RulesAction/DataConvert.php
```diff
@@ -43,13 +43,15 @@ class DataConvert extends RulesActionBase {
...
+   * @param string $rounding_behavior
+   *   The behaviour for rounding.
    */
-  public function execute() {
-    $value = $this->getContextValue('value');
-
-    $target_type = $this->getContextValue('target_type');
-    $rounding_behavior = $this->getContextValue('rounding_behavior');
-
+  public function doExecute($value, $target_type, $rounding_behavior) {
```
klausi: should be protected

src/Plugin/RulesAction/DataConvert.php
```diff
@@ -43,13 +43,15 @@ class DataConvert extends RulesActionBase {
 
   /**
    * Executes the plugin.
+   *
+   * @param mixed $value
+   *   The input value.
+   * @param string $target_type
+   *   The target for the action.
```
klausi: should be the "The target type the value should be converted into."

ndeet:
Fixed.

src/Plugin/RulesAction/DataConvert.php
```diff
@@ -43,13 +43,15 @@ class DataConvert extends RulesActionBase {
...
+   * @param string $rounding_behavior
+   *   The target type the value should be converted into.
```
klausi, line 50: nope, this is the rounding behavior.

ndeet:
Right, talking and coding does not work. Fixed.

klausi:
merged, thanks!

PR #308, opened 2015-11-29
2281119 Port user_has_role

ndf:
Renamed ->evaluate() to ->doEvaluate()


src/Plugin/Condition/UserHasRole.php
```diff
@@ -40,7 +40,7 @@ class UserHasRole extends RulesConditionBase {
   /**
    * {@inheritdoc}
    */
-  public function evaluate() {
+  protected function doEvaluate() {
     $account = $this->getContextValue('user');
```
klausi: this should be a method parameter now

src/Plugin/Condition/UserHasRole.php
```diff
@@ -40,7 +40,7 @@ class UserHasRole extends RulesConditionBase {
...
     $roles = $this->getContextValue('roles');
```
klausi: this should be a method parameter now

src/Plugin/Condition/UserHasRole.php
```diff
@@ -40,7 +40,7 @@ class UserHasRole extends RulesConditionBase {
   /**
    * {@inheritdoc}
```
klausi: this should document new method parameters.

src/Plugin/Condition/UserHasRole.php
```diff
@@ -38,12 +39,21 @@
...
+   *   Defaults to "AND".
+   *
+   * @return bool
+   *   TRUE if the account is blocked.
    */
-  public function evaluate() {
-    $account = $this->getContextValue('user');
-    $roles = $this->getContextValue('roles');
-    $operation = $this->getContext('operation')->hasContextValue() ? $this->getContextValue('operation') : 'AND';
+  protected function doEvaluate(UserInterface $account, array $roles, $operation = 'ALL') {
```
klausi: you mean AND instead of ALL?

src/Plugin/Condition/UserHasRole.php
```diff
@@ -38,12 +39,21 @@
...
+   *   The account to check.
+   * @param \Drupal\user\RoleInterface[] $roles
+   *   Array of user roles.
+   * @param string $operation
+   *   Either "AND": user has all of roles.
+   *   Or "OR": user has at least one of all roles.
+   *   Defaults to "AND".
+   *
+   * @return bool
+   *   TRUE if the account is blocked.
```
klausi: account is blocked? You mean TRUE if the user has the role(s)?

src/Plugin/Condition/UserHasRole.php
klausi: And can we add a "default:" clause to the switch() statement, so that AND applies if anything other than AND/OR is passed in?

src/Plugin/Condition/UserHasRole.php
ndf: Uhmm...

src/Plugin/Condition/UserHasRole.php
ndf: Should be AND offcourse.

Now throwing an InvalidArgumentException as fallback in the switch-statement.
This is also used in \Drupal\rules\Plugin\RulesAction\DataConvert

src/Plugin/Condition/UserHasRole.php
```diff
@@ -38,12 +39,22 @@
...
+   *   The account to check.
+   * @param \Drupal\user\RoleInterface[] $roles
+   *   Array of user roles.
+   * @param string $operation
+   *   Either "AND": user has all of roles.
+   *   Or "OR": user has at least one of all roles.
+   *   Defaults to "AND".
+   *
+   * @return bool
+   *   TRUE if user has role(s).
```
klausi: Should be "TRUE if the user has the role(s)."

src/Plugin/Condition/UserHasRole.php
```diff
@@ -38,12 +39,22 @@
...
+   *   Defaults to FALSE.
```
klausi: I think there is no "default" for return values, so this should be removed.

tests/src/Integration/Condition/UserHasRoleTest.php
```diff
@@ -88,4 +88,28 @@ public function testConditionEvaluation() {
...
+
+    $this->condition->setContextValue('user', $account->reveal());
+
+    $authenticated = $this->prophesize(RoleInterface::class);
+    $authenticated->id()->willReturn('authenticated');
+
+    // Now test INVALID as operation. An exception must be thrown.
+    $this->condition->setContextValue('roles', [$authenticated->reveal()]);
+    $this->condition->setContextValue('operation', 'INVALID');
+    $this->assertTrue($this->condition->evaluate());
```
klausi: you can remove the ->assertTrue() call, just execute evaluate() on the condition which will throw the exception already.

src/Plugin/Condition/UserHasRole.php
ndf: Removed the doc-comment and the "return FALSE" itself (it's an unreachable statement, says phpstorm)

klausi:
merged, thanks!

PR #311, opened 2015-11-30
Issue #2625986: enforce type string for plugin IDs.

torenware:
Drupal Issue https://www.drupal.org/node/2625986.

I'm starting out with bottleneck routines that get called a lot.  No tests yet, but I want to get this out there so people can figure out where we need to better enforce types.

No test yet.

### DIFF
--- a/src/Engine/ExpressionManager.php
+++ b/src/Engine/ExpressionManager.php
@@ -50,6 +50,9 @@ class ExpressionManager extends DefaultPluginManager implements ExpressionManage
    * {@inheritdoc}
    */
   public function createAction($id) {
+    if (!is_string($id)) {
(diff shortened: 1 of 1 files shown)


src/Engine/ExpressionManager.php
```diff
@@ -50,6 +50,9 @@ public function createReactionRule(array $configuration = []) {
    * {@inheritdoc}
    */
   public function createAction($id) {
+    if (!is_string($id)) {
+      throw new \InvalidArgumentException(t('Rules Action IDs must be strings.'));
```
klausi, line 54: Exception messages should not use t(), we are in the lowest levels of API here.

PR #315, opened 2015-12-04
class \Drupal\rules\Plugin\DataType\LoggerEntry

ndf:
See https://www.drupal.org/node/2625238
Started working on it


src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+namespace Drupal\rules\Plugin\DataType;
+
+use Drupal\Core\TypedData\Plugin\DataType\Map;
+
+/**
+ * The logger entry data type.
+ *
+ * @see: https://www.drupal.org/node/2625238
```
klausi: the "@see" annotations should not have a colon after it, see https://www.drupal.org/coding-standards/docs#see

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+ * @file
+ * Contains \Drupal\rules\Plugin\DataType\LoggerEntry.
+ */
+
+namespace Drupal\rules\Plugin\DataType;
+
+use Drupal\Core\TypedData\Plugin\DataType\Map;
+
+/**
+ * The logger entry data type.
```
klausi, line 24: Can we describe what this is? Example "Represents a log entry that is created from system log messages."

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+ * )
+ */
+class LoggerEntry extends Map {
+
+  /**
+   * The data definition.
+   *
+   * @var \Drupal\Core\TypedData\ComplexDataDefinitionInterface
+   */
+  protected $definition;
```
klausi: that property already exists on the parent class.

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * An array of values for the contained properties.
+   *
+   * @var array
+   */
+  protected $values = array();
```
klausi: that property already exists on the parent class.

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * The array of properties.
+   *
+   * @var \Drupal\Core\TypedData\TypedDataInterface[]
+   */
+  protected $properties = array();
```
klausi: that property already exists on the parent class.

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+
+  public function __construct(DataDefinitionInterface $definition, $name, TypedDataInterface $parent, $logger_entry_context) {
```
klausi: I think the constructor should not take the logger_entry array values, typed data is always set with the setValue() method.

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+    parent::__construct($definition, $name, $parent);
+
+    $this->values = $this->createLoggerEntryFromContext($logger_entry_context)
+  }
+
+  // Plain copy from \Drupal\rules\Logger\RulesLog
+  protected function createLoggerEntryFromContext($context) {
```
klausi: I think this method is not needed. The caller can just use ->setValue() and pass in the array. Almost all properties are primitives/strings except for variables.

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @ingroup typed_data
+ *
+ * @DataType(
+ *   id = "loggerentry",
+ *   label = @Translation("Logger Entry"),
+ *   definition_class = "\Drupal\Core\TypedData\MapDataDefinition"
+ * )
+ */
+class LoggerEntry extends Map {
```
klausi, line 34: you can look at examples for typed data in core: Language.php

src/Plugin/DataType/LoggerEntry.php
```diff
@@ -0,0 +1,106 @@
...
+ * The logger entry data type.
+ *
+ * @see: https://www.drupal.org/node/2625238
+ *
+ * @ingroup typed_data
+ *
+ * @DataType(
+ *   id = "loggerentry",
+ *   label = @Translation("Logger Entry"),
+ *   definition_class = "\Drupal\Rules\TypedData\LoggerEntryDataDefinition"
```
klausi: This should be our own class that extends MapDataDefinition.php and implements getPropertyDefinitions().

src/Logger/RulesLog.php
```diff
@@ -72,6 +73,10 @@ public function log($level, $message, array $context = array()) {
       'timestamp' => $context['timestamp'],
     );
 
+    $logger_entry = new LoggerEntry();
+    $logger_entry->log($level, $message, $context, $this->parser);
```
klausi, line 77: The logger entry should not have a log() method. It should have a setMessage() method, a setLevel() method etc.

config/schema/rules.schema.yml
```diff
@@ -205,3 +205,38 @@ rules.settings:
     log_level:
       type: integer
       label: 'Log level'
+
+rules.data_type.logger_entry.definition:
```
klausi, line 209: I don't think we need a config schema entry for the logger entry. It is not a config entity, it is just a typed data structure? Does the core Language data type for example have a config schema?

PR #318, opened 2015-12-07
Add API to delete expressions by ID from expression containers such as rules

klausi:
See https://www.drupal.org/node/2630412


src/Engine/ConditionExpressionContainer.php
```diff
@@ -139,4 +139,13 @@ public function getIterator() {
     return new \ArrayIterator($this->conditions);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function deleteExpressionAt($index) {
+    unset($this->conditions[$index]);
+    // Reshuffle array to refresh the indices.
+    $this->conditions = array_values($this->conditions);
```
fago: not sure about this - added my thoughts to the issue

src/Engine/ExpressionContainerInterface.php
```diff
@@ -48,4 +48,12 @@ public function addExpression($plugin_id, ContextConfig $config = NULL);
    */
   public function addExpressionObject(ExpressionInterface $expression);
 
+  /**
+   * Deletes an element indentified by the specified UUID in the container.
+   *
+   * @param string $uuid
+   *   The UUID of the element.
+   */
+  public function deleteExpression($uuid);
```
fago, line 73: This does not tell me whether deletion was successful?
I think we'll need a lookup by UUID also, so we can generate proper confirmation form with text etc.

src/Engine/ExpressionContainerInterface.php
```diff
@@ -48,4 +48,12 @@ public function addExpression($plugin_id, ContextConfig $config = NULL);
    */
   public function addExpressionObject(ExpressionInterface $expression);
 
+  /**
+   * Deletes an element indentified by the specified UUID in the container.
+   *
+   * @param string $uuid
+   *   The UUID of the element.
```
fago: expression

src/Engine/ExpressionContainerInterface.php
```diff
@@ -48,4 +48,12 @@ public function addExpression($plugin_id, ContextConfig $config = NULL);
    */
   public function addExpressionObject(ExpressionInterface $expression);
 
+  /**
+   * Deletes an element indentified by the specified UUID in the container.
```
fago: expression. Maybe removing from the container would be more suiting than "deleting" ?

src/Plugin/RulesExpression/ActionSet.php
```diff
@@ -77,7 +89,7 @@ public function addExpressionObject(ExpressionInterface $expression) {
     if (!$expression instanceof ActionExpressionInterface) {
       throw new InvalidExpressionException();
     }
-    $this->actions[] = $expression;
+    $this->actions[$this->uuidService->generate()] = $expression;
```
fago, line 93: Maybe we could make the manager responsible of generating IDs for expressions ? It seems a bit weird that every container has to take care of that. Likewise, I think we need a "global UUID lookup", which works at least on the configuration level as we do not know the container something is contaiend in either?

klausi, line 93: I don't think this belongs on the expression manager. It is just one call to generate a UUID, I would not couple it to the manger since it is none of the managers business where I get my UUIds from.

Implemented the lookup method as getExpression().

klausi:
merged.

PR #324, opened 2015-12-11
Feature/2632582

milkovsky:


tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -70,4 +82,85 @@ public function testFetchingByBasicPropertyPath() {
...
+  public function testFetchingEntityReference() {
+    $user = $this->entityTypeManager->getStorage('user')
+      ->create([
+        'name' => 'test',
+        'type' => 'user',
+      ]);
+    $this->node->uid->target_id = $user->id();
+
+    $fetched_user = $this->typedDataManager->getDataFetcher()
+      ->fetchByPropertyPath($this->node->getTypedData(), 'uid.entity.value')
```
fago: .value is non-sense, this does not exist. it should test "uid.0.entity" here

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -70,4 +82,85 @@ public function testFetchingByBasicPropertyPath() {
...
+      ->getValue();
+    $this->assertEquals(TRUE, $fetched_user instanceof EntityAdapter);
```
fago: asserTrue.
also, in this case use assertInstance UserInterface::class

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -70,4 +82,85 @@ public function testFetchingByBasicPropertyPath() {
...
+  }
+
+  /**
+   * @cover fetchByPropertyPath
+   */
+  public function testFetchingEntityReferenceAtPosition0() {
```
fago: position tests should use a multi-value field, that does not have to be a entity reference though. do not mix up things and clearly document what you are testing

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -70,4 +82,85 @@ public function testFetchingByBasicPropertyPath() {
...
+    $user2 = $this->entityTypeManager->getStorage('user')
+      ->create([
+        'name' => 'test2',
+        'type' => 'user',
+      ]);
+    $users[]['target_id'] = $user2->id();
+    $this->node->uid->setValue($users);
+
+    $fetched_user = $this->typedDataManager->getDataFetcher()
+      ->fetchByPropertyPath($this->node->getTypedData(), 'uid.1.entity.value')
```
fago: non-sense, uid is not multi-valued

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -70,4 +82,85 @@ public function testFetchingByBasicPropertyPath() {
...
+      ->getValue();
+    $this->assertEquals(TRUE, $fetched_user instanceof EntityAdapter);
+  }
+
+  /**
+   * @cover fetchByPropertyPath
+   */
+  public function testFetchingNonExistingEntityReference() {
+    $this->setExpectedException('Drupal\Core\TypedData\Exception\MissingDataException');
```
fago, line 204: usual there is some expectedException annotation, look at core examples or other rules examples

tests/src/Kernel/TypedData/DataFetcherTest.php
milkovsky: It tells about entityreference at position 0, so documented properly

tests/src/Kernel/TypedData/DataFetcherTest.php
milkovsky: where do I get this information from?

tests/src/Kernel/TypedData/DataFetcherTest.php
milkovsky: first you told EntityAdapter ?

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -46,16 +57,83 @@ public function setUp() {
...
       ->save();
 
-    $this->node = $entity_type_manager->getStorage('node')
+    $this->node = $this->entityTypeManager->getStorage('node')
       ->create([
         'title' => 'test',
         'type' => 'page',
       ]);
+
+    $this->installEntitySchema('entity_field_test');
```
fago, line 71: ? This installes a test entity - we don't need that, remove it. You can extend node?

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -46,16 +57,83 @@ public function setUp() {
...
+    $this->createTestField(
```
fago: nope, just create the field - see the example I linked you in the task

tests/src/Kernel/TypedData/DataFetcherTest.php
```diff
@@ -46,16 +57,83 @@ public function setUp() {
...
+   * @param string $field_name
+   *   The name of the field; if it already exists, a new instance of the existing
+   *   field will be created.
+   * @param string $field_label
+   *   The label of the field.
+   * @param string $target_entity_type
+   *   The type of the referenced entity.
+   * @param string $selection_handler
+   *   The selection handler used by this field.
+   * @param array $selection_handler_settings
```
fago, line 104: Why so many complicated parameters? I don't see why'd need to have a re-usable method at all, just create a test field in setup and we are done?

fago:
Merged manually.

PR #330, opened 2015-12-18
Implement the data selector / direct input mode switcher in the UI

klausi:
See https://www.drupal.org/node/2637102


klausi:
merged.

PR #336, opened 2016-01-13
Implement integrity checks for Rules expressions

klausi:
See https://www.drupal.org/node/2648352


src/Engine/ExpressionInterface.php
```diff
@@ -84,4 +84,17 @@ public function setConfigEntityId($id);
...
+   *
+   * Example: all variable names used in the expression are available.
+   *
+   * @param \Drupal\rules\Engine\ConfigurationStateInterface $config_state
+   *   The configuration state used to hold available data definitions of
+   *   variables.
+   *
+   * @return \Drupal\rules\Engine\IntegrityViolationList
+   */
+  public function integrityCheck(ConfigurationStateInterface $config_state);
```
fago: I think the method would be better named checkIntegrity()

src/Engine/ExpressionInterface.php
```diff
@@ -84,4 +84,17 @@ public function setConfigEntityId($id);
...
+  /**
+   * Verifies that this expression is configured correctly.
+   *
+   * Example: all variable names used in the expression are available.
+   *
+   * @param \Drupal\rules\Engine\ConfigurationStateInterface $config_state
+   *   The configuration state used to hold available data definitions of
+   *   variables.
+   *
+   * @return \Drupal\rules\Engine\IntegrityViolationList
```
fago: misses docs

src/Engine/ConfigurationState.php
```diff
@@ -0,0 +1,91 @@
...
+
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\Core\TypedData\DataDefinitionInterface;
+use Drupal\rules\Entity\ReactionRuleConfig;
+
+/**
+ * The state used during configuration time holding data definitions.
+ */
+class ConfigurationState implements ConfigurationStateInterface {
```
fago: Maybe "PreExecutionState" would be clearer? Else, it's hard to grasp what the state is used for.

src/Engine/ConfigurationState.php
```diff
@@ -0,0 +1,91 @@
...
+   * {@inheritdoc}
+   */
+  public static function create($data_definitions = []) {
+    return new static($data_definitions);
+    // @todo Initialize the global "site" variable.
+  }
+
+  public static function createFromConfig(ConfigEntityInterface $rules_config) {
+    $state = static::create();
+    if ($rules_config instanceof ReactionRuleConfig) {
```
fago: This is very much the same logic as we have to do during event dispatching on the execution state. Maybe we could centralize that part in the RulesComponent helper and add a method to get the configuration state from the helper object also?

src/Engine/ConfigurationStateInterface.php
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @param string $selector
+   *   The selector, example: "node:title:value".
+   * @param string $langcode
+   *   The langauge code.
+   *
+   * @return \Drupal\Core\TypedData\DataDefinitionInterface|null
+   *   A data definition if the selector could be applied, NULL otherwise.
+   */
+  public function applyDataSelector($selector, $langcode = LanguageInterface::LANGCODE_NOT_SPECIFIED);
```
fago: WE should move that logic to the DataFetcher

src/Engine/IntegrityViolation.php
```diff
@@ -0,0 +1,42 @@
...
+class IntegrityViolation {
+
+  protected $message;
+
+  protected $contextName;
+
+  protected $uuid;
+
+  public function setMessage($message) {
+    $this->message = $message;
```
fago, line 42: Do we need message translation here? I guess we could require the message to be pre-translated, can we?

src/Engine/IntegrityViolation.php
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public function getMessage() {
+    return $this->message;
+  }
+
+  public function getContextName() {
+    return $this->contextName;
```
fago: Not sure which context that is?

tests/src/Integration/Engine/IntegrityCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Tests that the integrity check can be invoked.
+   */
+  public function testIntegrityCheck() {
+    $rule = $this->rulesExpressionManager->createRule();
+    $rule->addAction('rules_entity_save', ContextConfig::create()
+      ->map('entity', 'entity')
+    );
+
+    $config_state = ConfigurationState::create([
```
fago: I think we should put this logic into the RulesComponent helper class also, such that you can do
RulesComponent($rule)
  ->checkIntegrity()
  ->execute();

(with execute being optional...)

tests/src/Integration/Engine/IntegrityCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+      'entity' => $this->typedDataManager->createDataDefinition('entity'),
+    ]);
+
+    $all_violations = $rule->integrityCheck($config_state);
+    $this->assertEquals(2, iterator_count($all_violations));
+
+    // Get the UUID of the second action.
+    $iterator = $rule->getIterator();
```
fago, line 71: ouch, I'd expect ArrayAccess to work here. Maybe this would be a good contributor task?

fago:
a good start - I added various comments :)

src/Engine/ConfigurationState.php
klausi: I'm not happy with "PreExecutionState" because it sounds like some state which is used before a rule is executed. But we are talking about config time here, which has nothing to do with execution.

src/Engine/ConfigurationState.php
klausi: yes, had the same idea. It is a similar logic to event execution, but different because this is configuration time. I will experiment a bit with places in RulesComponent and see how that looks.

src/Engine/IntegrityViolation.php
klausi, line 42: yes, this should accept string or TranslatableMarkup objects.

src/Engine/ConfigurationState.php
klausi: ExecutionMetadataState?

src/Engine/ConfigurationState.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Drupal\rules\Engine;
+
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\Core\TypedData\DataDefinitionInterface;
+use Drupal\rules\Entity\ReactionRuleConfig;
+
+/**
+ * The state used during configuration time holding data definitions.
```
klausi: Expand this comment.

klausi:
merged.

PR #338, opened 2016-01-14
Issue #2648918: Add new token replacement API (service).

fago:


src/TypedData/PlaceholderResolver.php
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+    $replacements = [];
+    $data_fetcher = $this->typedDataManager->getDataFetcher();
+    foreach ($placeholder_by_data as $data_name => $placeholders) {
+      foreach ($placeholders as $sub_path => $placeholder) {
+        try {
+          if (!isset($data[$data_name])) {
+            throw new MissingDataException();
```
klausi: can we have a message that describes what is missing where?

PR #340, opened 2016-01-16
Added composer.json

ModulesUnraveled:
I would love to be able to add the Rules module to my D8 projects using composer. Adding the composer.json file to the project root allows that, even during development.

For reference, the Redirect module is doing it here: https://github.com/md-systems/redirect

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,8 @@
+{
+  "name": "drupal/rules",
+  "description": "The Rules module allows site administrators to define conditionally executed actions based on occurring events.",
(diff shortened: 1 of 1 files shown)


klausi:
why is that necessary, that file is automatically added by drupal composer and Rules does not have any dependencies yet? https://packagist.drupal-composer.org/packages/drupal/rules#dev-8.x-3.x

ModulesUnraveled:
Ah, somehow I missed it when I looked at drupal-composer.org. Sorry. :(

PR #356, opened 2016-01-28
Test for init event

milkovsky:


milkovsky:
Work is in progress. Not ready to review

tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -157,4 +157,32 @@ public function testSystemLoggerEvent() {
...
+   */
+  public function testInitEvent() {
+    $rule = $this->expressionManager->createRule();
+    $rule->addCondition('rules_test_true');
+    $rule->addAction('rules_test_log');
+
+    $config_entity = $this->storage->create([
+      'id' => 'test_rule',
+      'expression_id' => 'rules_rule',
+      'event' => 'rules_system_logger_event',# @todo which event should we use?
```
milkovsky: which event should we use here?

tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -157,4 +157,32 @@ public function testSystemLoggerEvent() {
...
+    ]);
+    $config_entity->save();
+
+    // Rebuild the container so that the newly configured event gets picked up.
+    $this->container->get('kernel')->rebuildContainer();
+    // The logger instance has changed, refresh it.
+    $this->logger = $this->container->get('logger.channel.rules');
+
+    # @todo trigger Drupal init here.
+    #$this->container->get('cron')->run();
```
milkovsky: How to trigger Drupal initialization here?

milkovsky:
I have some issues with test. I do not understand so far how can we test Drupal initialization when it is already working.
The 'init' event is not a Rules event but kernel.request according to rules.rules.events.yml

tests/src/Kernel/EventIntegrationTest.php
fago: no way, just manually trigger the event somehow. e.g. get the dispatcher and dispatch

tests/src/Kernel/EventIntegrationTest.php
fago: the init event. look up the name in rules.rules_events.yml file

stephenpurkiss:
Here's some code which may be useful: https://www.drupal.org/node/2013014 has an example of creating a request, and links to https://www.drupal.org/sandbox/sam/2486229 which exposes as hook_init so shows how the symfony kernel.request event relates to the old hook_init rule.

HTH & thanks for picking this up!

milkovsky:
Thank you. Steve! As I understand we should only create a test, right?

stephenpurkiss:
Yes, because kernel.request is already defined as an event because it's just there - as the link above explains there is no hook_init, that is the equivalent of firing off a kernel.request which we can now do at any time as the example on that link shows with the redirect. So I am presuming we need to use code similar to that, or as Fago says dispatch, because that'll set off the event, if it doesn't then the test needs to fail. The gap in between is the code we need, which like you I hadn't figured out just yet but am sure you will, I need to carry on job hunting ;)

tests/src/Kernel/EventIntegrationTest.php
milkovsky: The init event does not exist. But there is kernel.request

stephenpurkiss: Please read the links I sent you, they explain that the init event does not exist and the comment on the link goes to a sandbox showing code on how to mimic the init event in 8.

stephenpurkiss: ...I.e. You need to test kernel.request event

milkovsky:
Thanks a lot, good luck :)

milkovsky:
Needs review

tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -157,4 +158,41 @@ public function testSystemLoggerEvent() {
...
+   */
+  public function testInitEvent() {
+    $rule = $this->expressionManager->createRule();
+    $rule->addCondition('rules_test_true');
+    $rule->addAction('rules_test_log');
+
+    $config_entity = $this->storage->create([
+      'id' => 'test_rule',
+      'expression_id' => 'rules_rule',
+      'event' => KernelEvents::REQUEST,
```
milkovsky, line 172: KernelEvents::REQUEST stands for 'kernel.request'

fago:
merged it.

PR #358, opened 2016-01-29
Issue #2632564 by fago: Make use of the new placeholder token replace…

fago:
…ment system.


tests/src/Kernel/CoreIntegrationTest.php
```diff
@@ -155,7 +155,7 @@ public function testTokenReplacements() {
       ->addContextDefinition('message', ContextDefinition::create('string'))
       ->addContextDefinition('type', ContextDefinition::create('string'))
       ->setContextValue('node', $node)
-      ->setContextValue('message', 'Hello [node:uid:entity:name:value]!')
+      ->setContextValue('message', 'Hello {{ node:uid:entity:name:value }}!')
```
klausi, line 158: Shouldn't this be converted to use "." instead of ":" separators?

fago:
already squashed and merged + fixed the separator

PR #361, opened 2016-01-29
Create the loop action expression plugin

klausi:
See https://www.drupal.org/node/2658960


src/Plugin/RulesExpression/RulesLoop.php
```diff
@@ -0,0 +1,93 @@
...
+use Drupal\rules\Engine\ExecutionMetadataStateInterface;
+use Drupal\rules\Engine\ExecutionStateInterface;
+use Drupal\rules\Engine\IntegrityViolationList;
+
+/**
+ * Holds a set of actions that are executed over the iteration of a list.
+ *
+ * @RulesExpression(
+ *   id = "rules_loop",
+ *   label = @Translation("Action set"),
```
fago: wrong label

src/Plugin/RulesExpression/RulesLoop.php
```diff
@@ -0,0 +1,93 @@
...
+ *   form_class = "\Drupal\rules\Form\Expression\ActionSetForm"
+ * )
+ */
+class RulesLoop extends ActionExpressionContainer {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function executeWithState(ExecutionStateInterface $state) {
+    $list_data = $state->getVariable($this->configuration['list']);
```
fago: The list needs to be configurable as any other context such that users can use the data selector here and property paths are supported. Would it make sense to make the loop plugin context aware so we can our context mapping and form traits for the context mapping?

src/Plugin/RulesExpression/RulesLoop.php
```diff
@@ -0,0 +1,93 @@
...
+
+    foreach ($list_data as $item) {
+      $state->setVariableData($list_item_name, $item);
+      foreach ($this->actions as $action) {
+        $action->executeWithState($state);
+      }
+    }
+    // After the loop the list item is out of scope and cannot be used by any
+    // following actions.
+    $state->deleteVariable($list_item_name);
```
fago: In d7 we cloned the state instead so we can leverage php scope handling for our scope. This fixes having a list item variable the overwrites a variable in the parent scope and makes the parent variable available again afterwards. Not sure we really support that, throwing an integrity exception in this case should be ok imo.
I think the d7 approach has some troubles with (open issues) of saving changes on the interim variables, so explictly removing them again + doing the integritycheck seems like a good approach to me. However "deleteVariable" sounds like it's being really deleted, can we call it remove? Also I think it would make sense to have a $save_changes parameter which defaults to TRUE.

tests/src/Integration/Engine/LoopTest.php
```diff
@@ -0,0 +1,261 @@
...
+    $rule->addExpressionObject($loop);
+
+    $violations = RulesComponent::create($rule)
+      ->addContextDefinition('string_list', ContextDefinition::create('string')->setMultiple())
+      ->addContextDefinition('existing_name', ContextDefinition::create('string'))
+      ->checkIntegrity();
+
+    $this->assertEquals(1, iterator_count($violations));
+    $this->assertEquals(
+      'List item name <em class="placeholder">existing_name</em> conflicts with an existing variable.',
```
fago, line 184: nice :)

src/Plugin/RulesExpression/RulesLoop.php
```diff
@@ -0,0 +1,93 @@
...
+    // If there are violations at this point stop checking here since it does
+    // not make sense to check the contained actions.
+    if (iterator_count($violation_list) > 0) {
+      return $violation_list;
+    }
+
+    $list_definition = $metadata_state->getDataDefinition($this->configuration['list']);
+    if ($list_definition instanceof ListDataDefinitionInterface) {
+      $list_item_definition = $list_definition->getItemDefinition();
+      $metadata_state->setDataDefinition($list_item_name, $list_item_definition);
```
fago, line 79: Interesting that this is necessary but makes sense. Might be nice to automate this via context api but I guess we'll see how things turn out to be easier when UI is added anyway.

klausi:
merged.

PR #364, opened 2016-01-30
Rebuild the container when a reaction rule is saved

chindris:
Rebuild the container when a reaction rule is saved and the generic subscriber is not already registered for the event of the rule.

d.o issue: https://www.drupal.org/node/2648284

### DIFF
--- a/src/Form/ReactionRuleEditForm.php
+++ b/src/Form/ReactionRuleEditForm.php
@@ -55,11 +55,11 @@ class ReactionRuleEditForm extends RulesComponentFormBase {
    * @param \Drupal\rules\Engine\RulesEventManager $event_manager
    *   The event plugin manager.
    */
-  public function __construct(RulesEventManager $event_manager, EventDispatcherInterface $eventDispatcher, EventSubscriberInterface $genericEventSubscriber, DrupalKernelInterface $drupalKernel) {
+  public function __construct(RulesEventManager $event_manager, EventDispatcherInterface $event_dispatcher, EventSubscriberInterface $generic_event_subscriber, DrupalKernelInterface $drupal_kernel) {
     $this->eventManager = $event_manager;
-    $this->eventDispatcher = $eventDispatcher;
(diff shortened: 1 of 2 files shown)


src/Form/ReactionRuleEditForm.php
```diff
@@ -95,4 +135,29 @@ protected function getRuleConfig() {
     return $this->entity;
   }
 
+  /**
+   * Checks if the event of the current rule is registered into the container.
+   *
+   * @return true | false
```
klausi: See https://www.drupal.org/coding-standards/docs#return on how to document return values.

klausi:
You also need to modify ConfigureAndExecuteTest where we do an extra cache clear button click that should not be necessary anymore with this change.

src/Form/ReactionRuleEditForm.php
```diff
@@ -95,4 +135,30 @@ protected function getRuleConfig() {
...
+    // in the listeners list for an object with the same class as our generic
+    // event subscriber which is registered for that event.
+    $event_name = $this->getRuleConfig()->getEvent();
+    $listeners = $this->eventDispatcher->getListeners();
+    $event_is_registered  = FALSE;
+    if (!empty($listeners[$event_name])) {
+      $generic_subscriber_class = get_class($this->genericEventSubscriber);
+      foreach ($listeners[$event_name] as $listener) {
+        if (is_object($listener[0]) && get_class($listener[0]) == $generic_subscriber_class) {
+          $event_is_registered = TRUE;
```
klausi: I think we don't need that variable, just return TRUE here?

src/Form/ReactionRuleEditForm.php
```diff
@@ -95,4 +135,30 @@ protected function getRuleConfig() {
...
+        }
+      }
+    }
+    return $event_is_registered;
```
klausi: And then this should be return FALSE

fago:
Squashed & merged.

PR #365, opened 2016-01-30
Add step-debugging guide for PHPStorm

Cogax:
I added a guide for debug tests in PHPStorm.


klausi:
Cool, can you add that to the rules book at https://github.com/fubhy/rules-docs ?

I think it is not really relevant for the brief README.

PR #368, opened 2016-01-30
EntityCreate context refining

Cogax:
I've made some refinings of the context from EntityCreate Action.
https://www.drupal.org/node/2658372

### DIFF
--- a/src/Plugin/RulesAction/EntityCreateDeriver.php
+++ b/src/Plugin/RulesAction/EntityCreateDeriver.php
@@ -89,14 +89,19 @@ class EntityCreateDeriver extends DeriverBase implements ContainerDeriverInterfa
           continue;
         }
 
-        $required = ($field_name == $bundle_key);
+        $isBundle = ($field_name == $bundle_key);
         $multiple = ($definition->getCardinality() === 1) ? FALSE : TRUE;
-        $this->derivatives["entity:$entity_type_id"]['context'][$field_name] = ContextDefinition::create($definition->getType())
-          ->setAssignmentRestriction(ContextDefinition::ASSIGNMENT_RESTRICTION_INPUT)
+        $contextDefinition = ContextDefinition::create($definition->getType())
           ->setLabel($definition->getLabel())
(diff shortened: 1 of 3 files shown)


Cogax:
When i want to make a new Action like "Create a new comment", I got an error:
`Recoverable fatal error: Argument 4 passed to Drupal\\rules\\Form\\Expression\\ActionForm::buildContextForm() must implement interface Drupal\\rules\\Context\\ContextDefinitionInterface, instance of Drupal\\Core\\Plugin\\Context\\ContextDefinition given, called in /modules/rules/src/Form/Expression/ActionForm.php on line 97 and defined in /modules/rules/src/Form/Expression/ContextFormTrait.php on line 21`

I don't know if thats assigned to my changes. Maybe you can help me? Thanks!

klausi:
Left a comment at https://www.drupal.org/node/2658372#comment-10804840

src/Plugin/RulesAction/EntityCreateDeriver.php
```diff
@@ -92,6 +92,7 @@ public function getDerivativeDefinitions($base_plugin_definition) {
         $required = ($field_name == $bundle_key);
         $multiple = ($definition->getCardinality() === 1) ? FALSE : TRUE;
         $this->derivatives["entity:$entity_type_id"]['context'][$field_name] = ContextDefinition::create($definition->getType())
+          ->setAssignmentRestriction(ContextDefinition::ASSIGNMENT_RESTRICTION_INPUT)
```
klausi: now you are adding the assignment restriction to every context definition, but you should only add it to the bundle context definition.

klausi: we should probably add test coverage for this. Add a second base field in the setUp() method of the test which is required. Add a third base field which is not required. Check that the generated derivative has 2 contexts, one for the bundle and one for the required base field. Check that the assignment restriction is only set on the bundle, not on the second field.

klausi:
merged, thanks!

PR #369, opened 2016-01-30
2654936 exceptions

emclaughlin1215:
See https://www.drupal.org/node/2654936

Here's my first pass on replacing generic exceptions with Rules-specific ones. It definitely needs more work, I'm more pull-requesting so someone will tell me what exactly more work needs to be done and offer guidance on how to continue the implementation.

Help?


src/Context/ContextConfig.php
```diff
@@ -60,7 +61,7 @@ protected function __construct(array $values) {
    * @param string $property_path
    *   A valid property path; e.g., "node.uid.target_id".
    *
-   * @throws \LogicException
+   * @throws \Drupal\rules\Exception\RulesLogicException
```
klausi: I think we don't need the "Rules" prefix for the exception name. LogicException in the rules namespace is enough.

src/Context/ContextDefinition.php
```diff
@@ -68,12 +69,15 @@ public function toArray() {
    * @param array $values
    *   The array of values, as returned by toArray().
    *
+   * @throws \Drupal\rules\Exception\RulesClassDefintionException
```
klausi: Throws docs should come after return docs.

emclaughlin1215:
The thing I am most unsure of in this update is the src/TypedData/DataFetcher.php changes. I'm not sure I'm handling throwing two exceptions with the same name from different classes right.

d7-tests/rules.test
```diff
@@ -191,8 +191,8 @@ class RulesTestCase extends DrupalWebTestCase {
       $action->execute();
       $this->fail("Execution hasn't created an exception.");
     }
-    catch (RulesEvaluationException $e) {
-      $this->pass("RulesEvaluationException was thrown: ". $e);
+    catch (EvaluationException $e) {
+      $this->pass("EvaluationException was thrown: ". $e);
```
klausi: This is the old Drupal 7 testing file and should NOT be modified. It is only there to keep track what we already cover and it is not executed.

src/Context/ContextDefinition.php
```diff
@@ -70,10 +71,13 @@ public function toArray() {
...
    * @return static
    *   The created definition.
+   *
+   * @throws \Drupal\rules\Exception\ClassDefintionException
+   *   If the required classes are not implemented.
    */
   public static function createFromArray($values) {
     if (isset($values['class']) && !in_array(ContextDefinitionInterface::class, class_implements($values['class']))) {
-      throw new \Exception('ContextDefinition class must implement ' . ContextDefinitionInterface::class . '.');
+      throw new ClassDefinitionException('ContextDefinition class must implement ' . ContextDefinitionInterface::class . '.');
```
klausi: Since this is a direct extension of the core context system I think we should use \Drupal\Component\Plugin\Exception\ContextException here.

src/Context/ContextProviderInterface.php
```diff
@@ -37,7 +37,7 @@ public function setProvidedValue($name, $value);
    * @param string $name
    *   The name of the provided context in the plugin definition.
    *
-   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   * @throws \Drupal\rules\Exception\ContextUndefinedInvalidException
```
klausi: I think we should not touch the exceptions that are already fully namespaced. This patch should only replace the global exceptions that don't have a namespace.

src/Context/ContextProviderTrait.php
```diff
@@ -54,7 +54,7 @@ public function getProvidedContext($name) {
   public function getProvidedContextDefinition($name) {
     $definition = $this->getPluginDefinition();
     if (empty($definition['provides'][$name])) {
-      throw new ContextException(sprintf("The %s provided context is not valid.", $name));
```
klausi, line 57: I think we should not touch the exceptions that are already fully namespaced. This patch should only replace the global exceptions that don't have a namespace.

src/Form/DeleteExpressionForm.php
```diff
@@ -10,7 +10,7 @@
 use Drupal\Core\Form\ConfirmFormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\rules\Entity\ReactionRuleConfig;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Drupal\rules\Exception\NotFoundHttpException;
```
klausi: I think we should not touch the exceptions that are already fully namespaced. This patch should only replace the global exceptions that don't have a namespace.

src/Engine/RulesComponent.php
```diff
@@ -208,9 +209,9 @@ public function execute() {
    * @return mixed[]
    *   The array of provided context values, keyed by context name.
    *
-   * @throws \LogicException
+   * @throws \Drupal\rules\Exception\L
```
klausi: incomplete?

emclaughlin1215:
It seems like the failing test is something that should be fixed in another branch, not this one, because it's only occurring in 8.1.x. Is that accurate? And if so, does that block this ticket/PR?

Cryt1c:
I have created a new pull request for this issue: https://github.com/fago/rules/pull/446

PR #372, opened 2016-01-31
Settings 2471693

DiDebru:


config/install/rules.settings.yml
```diff
@@ -1,4 +1,4 @@
-log_errors: warning
-debug_log: false
-debug: false
-log_level: info
+log_check: false
```
klausi, line 1: why "log_check"? Didn't we settle on just "log"?

config/install/rules.settings.yml
```diff
@@ -1,4 +1,4 @@
...
+debug_screen_log_check: false
```
klausi, line 2: Same here, can this just be "debug_screen"?

klausi:
tests are failing and coding standards also complain.

klausi:
continued in #404

PR #374, opened 2016-01-31
Move ConditionManager below the Core namespace and removed unused Rul…

fago:
…esEvent base class and interface.
See https://www.drupal.org/node/2660000


src/Core/RulesEventHandler.php
```diff
@@ -0,0 +1,36 @@
...
+    return !empty($definition['context']) ? $definition['context'] : [];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContextDefinition($name) {
+    $definition = $this->getPluginDefinition();
+    if (empty($definition['context'][$name])) {
+      throw new ContextException(sprintf("The %s context is not a valid context.", $name));
```
klausi: namespace for that exception is missing

klausi:
Did you forget to remove the old Condition/ConditionManager class?

klausi:
Did you forget to change RulesServiceProvider?

fago:
> Did you forget to change RulesServiceProvider?
> Nope, it's changed - it's just the use statement that changes.
> 
> Did you forget to remove the old Condition/ConditionManager class?
> Indeed, looks like I forgot to commit this - too much branch juggling ;/

fago:
fixe the points

src/Core/RulesConfigurableEventHandlerInterface.php
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Interface for handling configurable rules events.
+ *
+ * Configurable events have a custom event suffix, which gets appended to the
+ * base event name (= the plugin id), forming the fully-qualified event name,
+ * which is used for triggering reaction rules.
+ *
+ * For example, the fully-qualified event name of an event for viewing an
+ * article node would be "node_view--article", whereas "node_view" is the base
```
klausi: The event for viewing a node is "rules_entity_view:node", see rules_entity_view().

src/Core/RulesConfigurableEventHandlerInterface.php
```diff
@@ -0,0 +1,107 @@
...
+   *   The array of qualified event name suffixes to add; e.g, 'article' if
+   *   the fully-qualified event "node_view--article" should be triggered in
+   *   addition to base event "node_view".
+   */
+  public static function determineQualifiedEvents(Event $event, $event_name, array &$event_definition);
+
+  /**
+   * Provides a human readable summary of the event's configuration.
+   *
+   * @return string|\Drupal\Component\Render\MarkupInterface
```
klausi, line 55: Return description is missing.

src/Core/RulesDefaultEventHandler.php
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContextDefinition($name) {
+    $definition = $this->getPluginDefinition();
+    if (empty($definition['context'][$name])) {
+      throw new ContextException(sprintf("The %s context is not a valid context.", $name));
+    }
+    return $definition['context'][$name];
```
klausi: Why is there no refining of contexts here? Shouldn't this work the same as getContextDefinitions()? Please add a comment.

src/EventSubscriber/GenericEventSubscriber.php
```diff
@@ -111,11 +119,17 @@ public function onRulesEvent(Event $event, $event_name) {
...
+    // Invoke the rules.
+    foreach ($triggered_events as $triggered_event) {
+      // @todo Only load active reaction rules here.
+      $configs = $storage->loadByProperties(['event' => $triggered_event]);
+
+      // Loop over all rules and execute them.
+      foreach ($configs as $config) {
+        /** @var \Drupal\rules\Entity\ReactionRuleConfig $config */
+        $config->getExpression()
+          ->executeWithState($state);
```
klausi, line 134: Should we have a todo here that we should clone the execution state before executing the expression? Probably off-topic for this issue.

fago, line 134: Why do we want that? Such that added variables in a rule do not interact with another reaction rule? Makes sense, but that should probably be added with some dedicated test-coverage only.

src/Core/RulesConfigurableEventHandlerInterface.php
fago: :) Copied that part from d7

fago:
looks like github has issues with the squashed-merged branch, so added my fixes in https://github.com/fago/rules/pull/384

PR #375, opened 2016-01-31
Move RulesEventManager below the Core sub-namespace

fago:


klausi:
errr, I think the old title was correct?

klausi:
Looks good!

fago:
it's not finished ;)

fago:
ok, merged that.

PR #376, opened 2016-02-01
False integrity check violations on provided variables when they are used

klausi:
See https://www.drupal.org/node/2660204


klausi:
merged.

PR #377, opened 2016-02-01
addExpressionObject() method should always return $this

klausi:
See https://www.drupal.org/node/2659680


fago:
squashed and merged.

PR #380, opened 2016-02-02
Expression form errors

Cogax:
https://www.drupal.org/node/2659806


Cogax:
Not sure if I'm on the right way.
I'm working on test coverage..

klausi:
Looks like your 8.x-3.x branch is not up to date, in ActionForm there should be a limit_validation_errors key in the form.

src/Form/Expression/ActionForm.php
```diff
@@ -109,9 +109,18 @@ public function form(array $form, FormStateInterface $form_state) {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array $form, FormStateInterface $form_state) {
+    if ($form_state->get('action')) {
+      $this->submitForm($form, $form_state);
```
klausi, line 116: We should have a comment here why we call the submit handler here. Can be similar to the one we had in AddExpressionForm. And the comment should also talk about the first step of the form, which is the reson for the if() here.

src/Form/Expression/ConditionForm.php
```diff
@@ -111,9 +111,19 @@ public function form(array $form, FormStateInterface $form_state) {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array $form, FormStateInterface $form_state) {
+    if ($form_state->get('condition')) {
+      $this->submitForm($form, $form_state);
```
klausi, line 118: We should have a comment here why we call the submit handler here. Can be similar to the one we had in AddExpressionForm. And the comment should also talk about the first step of the form, which is the reason for the if() here.

klausi:
continued in #392

PR #381, opened 2016-02-03
Rebuild the container in the ReactionRuleStorage on update / delete and remove the rebuildContainer call from the tests.

chindris:
Three main things this patch does:
- moves the code that rebuilds the container from the Rule edit from into the ReactionRuleStorage class
- removes the calls to rebuildContainer() from the tests.
- rebuilds the container when needed also when a rule is deleted.

Issue on d.o: https://www.drupal.org/node/2659990


klausi:
this has been merged by fago.

PR #387, opened 2016-02-05
Prepare execution metadata state for an expression in the tree

klausi:
See https://www.drupal.org/node/2501939


src/Engine/ActionExpressionContainer.php
```diff
@@ -162,4 +162,30 @@ public function checkIntegrity(ExecutionMetadataStateInterface $metadata_state)
...
+  /**
+   * {@inheritdoc}
+   */
+  public function prepareExecutionMetadataState(ExecutionMetadataStateInterface $metadata_state, ExpressionInterface $until = NULL) {
+    if ($until) {
+      if ($this->getUuid() === $until->getUuid()) {
+        return TRUE;
+      }
+      foreach ($this->actions as $action) {
+        if ($action->getUuid() === $until->getUuid()) {
```
fago, line 174: not sure why this uses a ===, == should be fine.

src/Engine/ExpressionInterface.php
```diff
@@ -114,4 +114,25 @@ public function getUuid();
    */
   public function setUuid($uuid);
 
+  /**
+   * Prepares the execution metadata state.
```
fago: this docblock misses information on what "preparing" actually prepares. why would I use this method?

src/Engine/ExpressionInterface.php
```diff
@@ -114,4 +114,25 @@ public function getUuid();
...
+   *
+   * If this expression contains other expressions then the metadata state is
+   * set up recursively. If a $until expression is specified then the setup will
+   * stop right before that expression. This is useful for inspecting the state
+   * at a certain point in the expression tree, for example to do autocompletion
```
fago, line 123: this should explain what differs in the state

src/Engine/IntegrityCheckTrait.php
```diff
@@ -98,31 +98,24 @@ protected function doCheckIntegrity(CoreContextAwarePluginInterface $plugin, Exe
...
+            '%name' => $this->configuration['provides_mapping'][$name],
+          ]));
+          $violation->setContextName($name);
+          $violation->setUuid($this->getUuid());
+          $violation_list->add($violation);
         }
       }
     }
 
+    // Make sure that all proived variables by this plugin are added to the
```
fago: typo - provided

src/Engine/IntegrityCheckTrait.php
```diff
@@ -172,4 +165,33 @@ protected function checkDataTypeCompatible(ContextDefinitionInterface $context_d
...
 
+  /**
+   * Adds provided variables to the execution metadata state.
+   *
+   * @param CoreContextAwarePluginInterface $plugin
+   *   The action or condition plugin that may provide variables.
+   * @param \Drupal\rules\Engine\ExecutionMetadataStateInterface $metadata_state
+   *   The excution metadata state to add variables to.
+   */
+  public function doPrepareExecutionMetadataState(CoreContextAwarePluginInterface $plugin, ExecutionMetadataStateInterface $metadata_state) {
```
fago: doPrepareExecutionMetadataState() is a bad name imo as what preparing means differs by the plugin. but the trait only does "adding variables to it", so the name should be about that. e.g. addProvidedVariablesToExecutionMetadataState :)

src/Engine/RulesComponent.php
```diff
@@ -228,13 +228,23 @@ public function executeWithArguments(array $arguments) {
...
   public function checkIntegrity() {
+    $metadata_state = $this->getMetadataState();
+    return $this->expression->checkIntegrity($metadata_state);
+  }
+
+  /**
+   * Gets the metadata state.
+   *
+   * @return \Drupal\rules\Engine\ExecutionMetadataStateInterface
+   *   The populated execution metadata state.
```
fago: Should be clarified that it's the state at component invocation time imo. it does not describe the state after execution, but before execution.

src/Plugin/RulesExpression/RulesLoop.php
```diff
@@ -91,4 +92,46 @@ public function checkIntegrity(ExecutionMetadataStateInterface $metadata_state)
...
+  public function prepareExecutionMetadataState(ExecutionMetadataStateInterface $metadata_state, ExpressionInterface $until = NULL) {
+    if ($until && $this->getUuid() === $until->getUuid()) {
+      return TRUE;
+    }
+
+    $list_item_name = isset($this->configuration['list_item']) ? $this->configuration['list_item'] : 'list_item';
+    try {
+      $list_definition = $metadata_state->fetchDefinitionByPropertyPath($this->configuration['list']);
+      $list_item_definition = $list_definition->getItemDefinition();
+      $metadata_state->setDataDefinition($list_item_name, $list_item_definition);
```
fago, line 107: this adds the list item to the state even if $until is not passed. but it does not remove it again in this case. thus this needs to move into the if ($until) clause

tests/src/Integration/Engine/PrepareMetadataTest.php
```diff
@@ -0,0 +1,94 @@
...
+use Drupal\rules\Engine\ExecutionMetadataState;
+use Drupal\rules\Engine\RulesComponent;
+use Drupal\Tests\rules\Integration\RulesEntityIntegrationTestBase;
+
+/**
+ * Tests that the setup of the execution metadata state for an expression works.
+ *
+ * @group rules
+ */
+class PrepareMetadataTest extends RulesEntityIntegrationTestBase {
```
fago: PrepareExecutionMetadataStateTest :)

tests/src/Integration/Engine/PrepareMetadataTest.php
```diff
@@ -0,0 +1,94 @@
...
+  public function testPreparingUntil() {
+    // Setup a rule with 2 actions.
+    $rule = $this->rulesExpressionManager->createRule();
+    $rule->addAction('rules_variable_add', ContextConfig::create()
+      ->setValue('type', 'string')
+      ->setValue('value', '')
+      ->provideAs('variable_added', 'result1')
+    );
+    $second_action = $this->rulesExpressionManager->createAction('rules_variable_add')
+      ->setConfiguration(ContextConfig::create()
```
fago: It's not very nice that we have to do that just to get a reference on the added action :( Maybe we could add helper method to get the last expression added?  or could we just do end($rule->getActions()) ?

tests/src/Integration/Engine/PrepareMetadataTest.php
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public function testPreparingUntil() {
+    // Setup a rule with 2 actions.
+    $rule = $this->rulesExpressionManager->createRule();
+    $rule->addAction('rules_variable_add', ContextConfig::create()
+      ->setValue('type', 'string')
+      ->setValue('value', '')
+      ->provideAs('variable_added', 'result1')
+    );
+    $second_action = $this->rulesExpressionManager->createAction('rules_variable_add')
```
fago: unrelated and minor: but looks like createAction misses a second argument with ContextConfig as else there is no nice API set it? or we provide an alternate, nicer setter in addition to setConfiguation() ?

tests/src/Integration/Engine/PrepareMetadataTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+    $rule->addExpressionObject($loop);
+
+    $state = RulesComponent::create($rule)
+      ->addContextDefinition('string_list', ContextDefinition::create('string')->setMultiple())
+      ->getMetadataState();
+
+    $found = $rule->prepareExecutionMetadataState($state, $action);
+    $this->assertTrue($state->hasDataDefinition('list_item'));
+    $this->assertTrue($found);
```
fago: This looks good, but misses coverage that after the loop the list item is gone.

Also, I'm wondering whether it would be a good idea to have a LoopExpressionTest case which agggregates all loop related test-coverage. Thus, there could be one method for every aspect to test like the metadata and execution time test coverage. Then further expressions like and and or plugins can follow the same pattern?

src/Engine/IntegrityCheckTrait.php
```diff
@@ -98,31 +98,24 @@ protected function doCheckIntegrity(CoreContextAwarePluginInterface $plugin, Exe
...
+          $violation->setContextName($name);
+          $violation->setUuid($this->getUuid());
+          $violation_list->add($violation);
         }
       }
     }
 
+    // Make sure that all proived variables by this plugin are added to the
+    // execution metadata state.
+    $this->doPrepareExecutionMetadataState($plugin, $metadata_state);
```
fago: oh I see now that the plugin is required to update that method with suiting state adaptions as necessary - however that's not obvious for the caller / wasn't obvious for me :) Thus this needs to be documented.
Besides that, I wonder whether that would not belong into ContextHandlerTrait really? The trait cares already of the context mapping for execution time, so would it make sense to care about at configuration time as well? Or if we want to separate it for better code organization, we could make a ExecutionContextHandlerTrait & a ExecutionMetadataHandlerTrait below the Context namespace?

src/Engine/ActionExpressionContainer.php
klausi, line 174: "===" is the sane default when comparing stuff that must be identical, so I would leave that as is.

src/Engine/ExpressionInterface.php
klausi, line 123: done

src/Plugin/RulesExpression/RulesLoop.php
klausi, line 107: The list item is removed in both cases, so this works as designed. Added a test case to also make sure of that.

tests/src/Integration/Engine/PrepareMetadataTest.php
klausi: no, end() does not work with iterators. It would have to be "$rule->getIterator()->offsetGet(1);" which is also not very nice. For conditions in a rule "$rule->getConditions()->getIterator()->offsetGet(1);".

I'm leaving this as is for now.

tests/src/Integration/Engine/PrepareMetadataTest.php
klausi: createAction() on the expression manager takes an array, same as createRule() does. We could change that interface to take ContextConfig instead in a separate issue?

tests/src/Integration/Engine/PrepareMetadataTest.php
klausi: Test coverage for that list item after the loop has been added.

We have a test case for loops (LoopTest), so we could add it there. Currently the tests are arranged by topic (integrity check, auto saving, RulesComponent usage), but also per expression (LoopTest and the unit tests for actions sets and condition sets).

I'm not sure which is better, since only one test class for an expression could also grow quite a lot - I like the topical tests.

Leaving this as is for now.

klausi:
merged.

PR #390, opened 2016-02-05
Port mail_to_users_of_role action to D8 (continues PR #217)

super-romeo:
Following up from the work done on [#217](https://github.com/fago/rules/pull/217).

### DIFF
--- a/src/Plugin/RulesAction/SystemMailToUsersOfRole.php
+++ b/src/Plugin/RulesAction/SystemMailToUsersOfRole.php
@@ -1,0 +1,156 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Plugin\RulesAction\SystemMailToUsersOfRole.
+ */
+
+namespace Drupal\rules\Plugin\RulesAction;
+
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Mail\MailManagerInterface;
+use Drupal\rules\Core\RulesActionBase;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Config\ImmutableConfig;
+use Drupal\user\UserStorage;
+
+/**
+ * Provides a 'Mail to users of a role' action.
+ *
+ * @RulesAction(
+ *   id = "rules_mail_to_users_of_role",
+ *   label = @Translation("Mail to users of a role"),
+ *   category = @Translation("System"),
+ *   context = {
+ *     "roles" = @ContextDefinition("entity:user_role",
+ *       label = @Translation("Roles"),
+ *       description = @Translation("The roles to which to send the e-mail."),
+ *       multiple = TRUE
+ *     ),
+ *     "subject" = @ContextDefinition("string",
+ *       label = @Translation("Subject"),
+ *       description = @Translation("The subject of the e-mail."),
+ *     ),
+ *     "body" = @ContextDefinition("string",
+ *       label = @Translation("Body"),
+ *       description = @Translation("The body of the e-mail."),
+ *     ),
+ *     "from" = @ContextDefinition("email",
+ *       label = @Translation("From"),
+ *       description = @Translation("The from e-mail address."),
+ *       required = FALSE
+ *     )
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ */
+class SystemMailToUsersOfRole extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var \Psr\Log\LoggerInterface
+   */
+  protected $logger;
+
+  /**
+   * @var \Drupal\Core\Mail\MailManagerInterface
+   */
+  protected $mailManager;
+
+  /**
+   * @var \Drupal\Core\Config\ImmutableConfig
+   */
+  protected $config;
+
+  /**
+   * @var \Drupal\user\UserStorage
+   */
+  protected $userStorage;
+
+  /**
+   * Constructs a SendMailToUsersOfRole object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Psr\Log\LoggerInterface $logger
+   *   The storage service.
+   * @param \Drupal\Core\Mail\MailManagerInterface $mail_manager
+   *   The mail manager service.
+   * @param \Drupal\Core\Config\ImmutableConfig $site_config
+   *   The site configuration.
+   * @param \Drupal\user\UserStorage $userStorage
+   *   The user storage service.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, LoggerInterface $logger, MailManagerInterface $mail_manager, ImmutableConfig $site_config, UserStorage $userStorage) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->logger = $logger;
+    $this->mailManager = $mail_manager;
+    $this->config = $site_config;
+    $this->userStorage = $userStorage;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('logger.factory')->get('rules'),
+      $container->get('plugin.manager.mail'),
+      \Drupal::config('system.site'),
+      $container->get('entity.manager')->getStorage('user')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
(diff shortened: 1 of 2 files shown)


tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,220 @@
...
+  protected $action;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->enableModule('user');
+    // @todo why is that line needed? Please add a comment.
+    $this->namespaces[] = '';
```
klausi: So is this line still needed? Did you try without that line? What was the error in that case?

klausi:
test are failing at https://travis-ci.org/fago/rules/jobs/107259039 , did you try to run the test locally? Did they work?

klausi:
ah, and there are some coding standard violations, can you also fix those as reported by Travis CI?

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,156 @@
...
+ *   }
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
+ */
+class SystemMailToUsersOfRole extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var \Psr\Log\LoggerInterface
+   */
```
klausi, line 56: Can we add docs to all this class variables? See https://www.drupal.org/node/1354#var

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,156 @@
...
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('logger.factory')->get('rules'),
+      $container->get('plugin.manager.mail'),
+      \Drupal::config('system.site'),
```
klausi: don't use \Drupal in classes, use $container that you get here instead.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,156 @@
...
+      $container->get('entity.manager')->getStorage('user')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
```
klausi: we should use the doExceute() pattern here as done by other actions.

superromeo: I will fix all this problems tomorrow morning.
And btw D8 is great! I like it :)
Thank you for help!
5 фев 2016 г. 18:52 пользователь Klaus Purer <notifications@github.com> написал:In src/Plugin/RulesAction/SystemMailToUsersOfRole.php:
> +  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
> +    return new static(
> +      $configuration,
> +      $plugin_id,
> +      $plugin_definition,
> +      $container->get('logger.factory')->get('rules'),
> +      $container->get('plugin.manager.mail'),
> +      \Drupal::config('system.site'),
> +      $container->get('entity.manager')->getStorage('user')
> +    );
> +  }
> +
> +  /**
> +   * {@inheritdoc}
> +   */
> +  public function execute() {

we should use the doExceute() pattern here as done by other actions.

super-romeo:
With entity:user_role I get one error for each test:
"
Drupal\Component\Plugin\Exception\PluginNotFoundException: The "entity:user_role" plugin does not exist.

/var/www/test_8/web/core/lib/Drupal/Component/Plugin/Discovery/DiscoveryTrait.php:57
/var/www/test_8/web/core/lib/Drupal/Component/Plugin/Discovery/DiscoveryCachedTrait.php:30
/var/www/test_8/web/core/lib/Drupal/Core/TypedData/TypedDataManager.php:132
/var/www/test_8/web/core/lib/Drupal/Core/Plugin/Context/ContextDefinition.php:240
/var/www/test_8/web/core/lib/Drupal/Core/Plugin/Context/Context.php:106
/var/www/test_8/web/core/lib/Drupal/Core/Plugin/Context/Context.php:57
/var/www/test_8/web/core/lib/Drupal/Core/Plugin/Context/Context.php:179
/var/www/test_8/web/core/lib/Drupal/Core/Plugin/ContextAwarePluginBase.php:78
/var/www/_test_task/rules/tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php:222
/var/www/_test_task/rules/tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php:171
"

Just like was here: https://www.drupal.org/node/2409059#comment-10092104.

UserRoleAdd and UserRoleRemove uses entity:user_role as context and it works fine. I could not find what causes the problem here. Any idea?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    $this->logger = $this->prophesize(LoggerInterface::class);
+    $logger_factory = $this->prophesize(LoggerChannelFactoryInterface::class);
+    $logger_factory->get('rules')->willReturn($this->logger->reveal());
+
+    $this->mailManager = $this->prophesize(MailManagerInterface::class);
+    $this->userStorage = $this->prophesize(UserStorageInterface::class);
+
+    // Mocked entityManager.
+    $this->entityManager = $this->prophesize(EntityManagerInterface::class);
```
klausi, line 115: The entity manager is already mocked in the parent class, so this line is not necessary.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,239 @@
...
+    $this->entityManager->getStorage('user')
+      ->willReturn($this->userStorage->reveal());
+
+    $this->entityManager->createInstance()
+      ->willReturn($this->userStorage->reveal());
+
+    // Prepare a user entity type instance.
+    $entity_type = reset($this->users)->getEntityType();
+    $this->entityManager->getDefinitions()
+      ->willReturn($entity_type);
```
klausi, line 126: Problem: now the entity manager only returns the definition for the user entity type, but not for the user role entity type. I think this is not necessary because the parent class RulesEntityIntegrationTestBase already mocks that.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+      // Mock the role entity.
+      $role_mock = $this->prophesize(RoleInterface::class);
+      $role_mock->id()
+        ->willReturn($role_name);
+      $this->roles[] = $role_mock->reveal();
+
+      // Mock the user entity.
+      $user_mock = $this->prophesizeEntity(UserInterface::class);
+      $user_mock->id()->willReturn($uid);
```
klausi, line 91: Now both users will have the uid 1. I think they should return different uids.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,239 @@
...
+   * @param int $call_number
+   *   The number of emails that should be sent.
+   * @param int $role_number
+   *   The number of roles (1 or 2).
+   *
+   * @dataProvider providerSendMailToRole
+   *
+   * @covers ::execute
+   */
+  public function testSendMailToRole($call_number, $role_number) {
```
klausi, line 163: Instead of passing in $role_number here you should pass in the actual role object.

klausi, line 163: Can we rename $call_number to $email_send_count ?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,239 @@
...
+    // Unfortunately providerSendMailToRole() runs before setUp() so we can't
+    // set these things up there.
+    // Sending mail to one role.
+    if ($role_number == 1) {
+      $user = reset($this->users);
+      $roles = [$this->roles[0]];
+      $users = [$user->id() => $user];
+      $this->helperSendMailToRole($call_number, $roles, $users);
```
klausi, line 171: I don't think you need that helper method. Can we bring that directly into this method? after the if()?

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,173 @@
...
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('logger.factory')->get('rules'),
+      $container->get('plugin.manager.mail'),
+      $container->get('config.factory')->get('system.site'),
+      $container->get('entity.manager')->getStorage('user')
```
klausi, line 117: we shouldn't use the entity.manager because it is deprecated. Use entity type manager instead.

klausi:
continued in #403

PR #396, opened 2016-02-07
Convert RulesUI to work based on RulesComponent

fago:
See https://www.drupal.org/node/2659028

WIP


src/Entity/ReactionRuleConfig.php
```diff
@@ -160,6 +160,36 @@ public function getExpression() {
...
+
+  /**
+   * Sets the Rules component containing the configuration to be stored.
+   *
+   * @param \Drupal\rules\Engine\RulesComponent $component
+   *   The component.
+   *
+   * @return $this
+   */
+  public function setComponent(RulesComponent $component) {
```
klausi: setComponent()? But no component is set in the implementation? Should this be setComponentConfiguration() or transferComponentConfig() or similar?

fago: true, the name might be misleading - maybe setByComponent ?

klausi:
looks like you mixed up the pull request and pushed some unrelated changes?

klausi:
Aha, maybe this is the correct pull request, fixed the title and added a link to the d.o issue.

rules.services.yml
```diff
@@ -34,8 +34,13 @@ services:
...
     tags:
       - { name: 'event_subscriber' }
+  rules.ui_route_enhancer:
+    class: Drupal\rules\Routing\RulesUiRouteEnhancer
+    arguments: ['@plugin.manager.rules_ui']
+    tags:
+      - { name: route_enhancer }
   rules.ui_route_subscriber:
-    class: Drupal\rules\Routing\UiRouteSubscriber
+    class: Drupal\rules\Routing\RulesUiRouteSubscriber
```
klausi, line 43: why the name change? No need for the "Rules" prefix?

src/Entity/ReactionRuleConfig.php
klausi: The point is that config is transfered over, so should "config" be in the name?

rules.services.yml
fago, line 43: it's a bit more consistent with the other routing class (enhancer), the RulesUI term and eases finding the subscriber when doing class lockups. I stumbled over it multiple times.

fago:
That's merged.

PR #398, opened 2016-02-07
Enabled DrupalPractice PHPCS checking

klausi:


klausi:
Meh, that would require the dev version of Coder ... let's wait for the next Coder release with this.

PR #399, opened 2016-02-08
Expose Rules components as action/condition plugin

klausi:
See https://www.drupal.org/node/2664700


src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Provides a generic 'Execute Rules component' action.
+ *
+ * @RulesAction(
+ *   id = "rules_component",
+ *   deriver = "Drupal\rules\Plugin\RulesAction\RulesComponentActionDeriver"
+ * )
+ *
+ * @todo: Add access callback information from Drupal 7.
```
fago, line 23: we have configure access and access() so I think this should be doable

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,78 @@
...
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $storage;
+
+  /**
+   * The ID of the rules component config entity.
+   *
+   * @var string
+   */
+  protected $componentId;
```
fago, line 39: this should be part of the action definition

tests/src/Integration/Engine/ComponentActionTest.php
```diff
@@ -0,0 +1,42 @@
...
+  public function testActionAvailable() {
+    $rule = $this->rulesExpressionManager->createRule();
+
+    $rules_config = new RulesComponentConfig([
+      'id' => 'test_rule',
+      'label' => 'Test rule',
+    ], 'rules_component');
+    $rules_config->setExpression($rule);
+
+    $storage = $this->prophesize(ConfigEntityStorageInterface::class);
```
fago: will this is fine / nice to mock now, I think we'll need a test case including the config system in the end as we'll have to cover configuration system dependencies. Thus, we either split that in future or just do a single kernel test case already now.

src/Plugin/RulesAction/RulesComponentAction.php
klausi, line 39: I don't think so because we expose the component ID already in the action plugin defintion. The derivative should not expose the component ID as context definition.

Or do you think we should not implement this with derivatives? We could also have only once action plugin that takes the component ID as context.

tests/src/Integration/Engine/ComponentActionTest.php
klausi: We can add an additional kernel test for config dependencies, but for testing derivative and context definition functionality this integrity test should be fine and is much faster to execute.

src/Plugin/RulesAction/RulesComponentAction.php
klausi, line 23: Doable how? We don't have https://www.drupal.org/node/2418165 yet.

fago, line 23: hm, it does not look so - since we moved away from core actions we lost support for access() also. Looks like we need to work on that.

src/Plugin/RulesAction/RulesComponentAction.php
fago, line 39: no, you are right. I just seemed duplicating as it already part of the the plugin definition. I guess it's a short-cut? It's a bit duplicating data now, so a method would be nicer - but no biggie.

src/Engine/ExecutionStateInterface.php
```diff
@@ -122,6 +122,15 @@ public function fetchDataByPropertyPath($property_path, $langcode = NULL);
   public function saveChangesLater($selector);
 
   /**
+   * Returns the list of variables that should be auto saved after execution.
+   *
+   * @return string[]
+   *   The list of data selectors that specify the target object to be saved.
+   *   Example: node:uid:entity.
```
fago: Wrong example, we use points now as delimiter.

src/Engine/ExecutionStateInterface.php
```diff
@@ -122,6 +122,15 @@ public function fetchDataByPropertyPath($property_path, $langcode = NULL);
   public function saveChangesLater($selector);
 
   /**
+   * Returns the list of variables that should be auto saved after execution.
```
fago: minor: auto-saved ?

src/Plugin/RulesAction/EntityCreate.php
```diff
@@ -75,7 +75,9 @@ public function refineContextDefinitions() {
     if ($type = $this->entityTypeId) {
       $data_type = "entity:$type";
 
-      if ($bundle = $this->getContextValue('bundle')) {
+      // @todo fix this to get the proper bundle context name, do not hardcode
```
fago: Let's create an issue for that and reference it here.

src/Plugin/RulesAction/EntityCreateDeriver.php
```diff
@@ -89,9 +89,19 @@ public function getDerivativeDefinitions($base_plugin_definition) {
           continue;
         }
 
+        // If this is an entity reference then we expect the target type as
+        // context.
+        if ($definition->getType() === 'entity_reference') {
+          $type = 'entity:' . $definition->getSetting('target_type');
```
fago: This hard-codes two things it shouldn't / hasn't:
- entity_references do not have to be of type 'entity_reference'. Let's check for a data reference interface instead.
- This should not hard-code the setting-name, which is internal to a specific field type. It should use the getTargetDefintion (or how it's called) and call getDataType() on it.

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,132 @@
...
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $rules_config = $this->storage->load($this->componentId);
+
+    // Setup an isolated execution state for this expression and pass on the
+    // necessary context.
+    $rules_component = $rules_config->getComponent();
+    foreach ($this->getContextValues() as $context_name => $context_value) {
+      $rules_component->setContextValue($context_name, $context_value);
```
fago: I think this should better use getState()->setVariableData() and pass on the context Data object - so we avoid re-creating it.

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,132 @@
...
+    foreach ($state->getAutoSaveSelectors() as $selector) {
+      $parts = explode('.', $selector);
+      $context_name = reset($parts);
+      if (array_key_exists($context_name, $this->context)) {
+        $this->saveLater[] = $context_name;
+      }
+      else {
+        // Otherwise we need to save here since it will not happen in the parent
+        // execution.
+        $typed_data = $state->fetchDataByPropertyPath($selector);
```
fago: This should get the variable from the state instead. For one it's already there, for second it might have been changed somehow?

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,132 @@
...
+          // Saving is always done at the root of the typed data tree, for example
+          // on the entity level.
+          $typed_data->getRoot()->getValue()->save();
+        }
+      }
+    }
+
+    $provided_context = $rules_component->getProvidedContext();
+    foreach ($provided_context as $name) {
+      $this->setProvidedValue($name, $state->getVariableValue($name));
```
fago: Again, let's rather pass through the typed data object?
-> get the provided context object and set the variable via setContextData().

src/Plugin/RulesAction/RulesComponentAction.php
klausi: what do you mean? We are fetching from the state here?

config/schema/rules.action.schema.yml
```diff
@@ -1,3 +1,13 @@
+# For undefined context values just pretend this is a list of strings.
+# We need this to make
+# Drupal\Tests\rules\Kernel\CoreIntegrationTest::testComponentActionAutoSave
+# work.
+rules.action.context_values.*:
+  type: sequence
+  label: Context values
+  sequence:
+    type: string
```
fago: Wouldn't it be better to use the type "ignore" then and make it officially optional to have something better?

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,133 @@
...
+
+    // We don't use RulesComponent::execute() here since we don't want to
+    // auto-save immediately.
+    $state = $rules_component->getState();
+    $expression = $rules_component->getExpression();
+    $expression->executeWithState($state);
+
+    // Postpone auto-saving to the parent expression triggering this action.
+    foreach ($state->getAutoSaveSelectors() as $selector) {
+      $parts = explode('.', $selector);
```
fago, line 101: This looks like auto-saving on selected properties will not work. I guess, that might be a general issue? I'd suggest opening a follow-up for that.

fago, line 101: Probably we need a helper in the data fetcher which gets the selector of the parent root item, i.e. find the parent DataReference or root.

src/Plugin/RulesAction/RulesComponentAction.php
fago: right, nevermind!

src/Plugin/RulesAction/RulesComponentAction.php
```diff
@@ -0,0 +1,133 @@
...
+      if (array_key_exists($context_name, $this->context)) {
+        $this->saveLater[] = $context_name;
+      }
+      else {
+        // Otherwise we need to save here since it will not happen in the parent
+        // execution.
+        $typed_data = $state->fetchDataByPropertyPath($selector);
+        // The returned data can be NULL, only save it if we actually have
+        // something here.
+        if ($typed_data) {
```
fago: If it's null there should be still a typed data object wrapping it, so I guess this should check whether the content is NULL. But is that actually a valid case? I mean a plugin telling us to save something that is NULL sounds like a bug in the plugin.

tests/src/Integration/Action/EntityCreateTest.php
```diff
@@ -43,23 +46,33 @@ public function setUp() {
...
-    $bundle_field_definition->getType()->willReturn('string')
-      ->shouldBeCalledTimes(1);
+    $bundle_field_definition->getType()->willReturn('string');
     $bundle_field_definition->getLabel()->willReturn('Bundle')
       ->shouldBeCalledTimes(1);
     $bundle_field_definition->getDescription()
       ->willReturn('Bundle mock description')
       ->shouldBeCalledTimes(1);
 
+    $bundle_field_definition_required->getItemDefinition()
```
fago, line 71: I dislike this test-case, maintainability--  Too much mocking going on here. But that's pre-existing, just mentioning.

tests/src/Integration/Action/RulesComponentActionTest.php
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Tests that auto saving is only triggered once with nested components.
+   */
+  public function testAutosaveOnlyOnce() {
+    $entity = $this->prophesizeEntity(EntityInterface::class);
+
+    $nested_rule = $this->rulesExpressionManager->createRule();
+    $nested_rule->addAction('rules_data_set', ContextConfig::create()
+      ->map('data', 'entity')
```
fago: I think the test coverage is bogus as setting an "entity" variable with a new entity should NOT trigger auto-saving at all. It should only save the entity if you change it, not if you set some variable with it. It might have to save the parent data structure though, but that's another issue. I think the test case should just use the entity save action.

klausi:
merged.

PR #402, opened 2016-02-09
Add provided context definitions to Rules component config

klausi:
See https://www.drupal.org/node/2665264

### DIFF
--- a/src/Entity/RulesComponentConfig.php
+++ b/src/Entity/RulesComponentConfig.php
@@ -41,6 +41,7 @@ use Drupal\rules\Engine\ExpressionInterface;
  *     "core",
  *     "expression_id",
  *     "context_definitions",
+ *     "provided_context_definitions",
  *     "configuration",
  *   },
  *   links = {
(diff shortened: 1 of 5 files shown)


src/Entity/RulesComponentConfig.php
```diff
@@ -214,6 +221,36 @@ public function setContextDefinitions($definitions) {
   }
 
   /**
+   * Gets the provided definitions of the used context.
```
fago: provided or used? I think "used" should be removed here.

src/Entity/RulesComponentConfig.php
```diff
@@ -214,6 +221,36 @@ public function setContextDefinitions($definitions) {
...
+  public function getProvidedContextDefinitions() {
+    $definitions = [];
+    foreach ($this->provided_context_definitions as $name => $definition) {
+      $definitions[$name] = ContextDefinition::createFromArray($definition);
+    }
+    return $definitions;
+  }
+
+  /**
+   * Sets the provided definitions of the used context.
```
fago: again, used or provided?

klausi:
merged.

PR #405, opened 2016-02-11
Change reaction-rule config to save the event with settings

milkovsky:
https://www.drupal.org/node/2666200


config/schema/rules.schema.yml
```diff
@@ -194,9 +206,21 @@ rules_expression.rules_reaction_rule:
     uuid:
       type: string
       label: 'UUID'
-    event:
-      type: string
-      label: 'Event'
+    events:
```
fago: unrelated, this seems wrong  / outdated. Probably we can remove rules_expression.rules_reaction_rule - the plugin does not exist any more afair.

src/Entity/ReactionRuleConfig.php
```diff
@@ -121,11 +121,15 @@ class ReactionRuleConfig extends ConfigEntityBase {
   protected $module = 'rules';
 
   /**
-   * The event name this reaction rule is reacting on.
+   * The events this reaction rule is reacting on.
    *
-   * @var string
+   * Events array, key - numeric index, value - event array with next structure:
```
fago: Should be a full sentence alway:
The array is numerically indexed and contains arrays with the following structure:

tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -185,4 +185,38 @@ public function testInitEvent() {
...
+    $config_entity = $this->storage->create([
+      'id' => 'test_rule',
+      'expression_id' => 'rules_rule',
+      'events' => [
+        ['event_name' => 'rules_user_login'],
+        ['event_name' => 'rules_user_logout'],
+      ],
+      'configuration' => $rule->getConfiguration(),
+    ]);
+    $config_entity->save();
```
fago, line 199: This should test using the getter also.

tests/src/Kernel/EventIntegrationTest.php
```diff
@@ -185,4 +185,38 @@ public function testInitEvent() {
...
+
+    // The logger instance has changed, refresh it.
+    $this->logger = $this->container->get('logger.channel.rules');
+
+    $account = User::create(['name' => 'test_user']);
+    // Invoke the hook manually which should trigger the rules_user_login event.
+    rules_user_login($account);
+    // Invoke the hook manually which should trigger the rules_user_logout
+    // event.
+    rules_user_logout($account);
```
fago, line 209: Nice that this works already!

src/Entity/ReactionRuleConfig.php
```diff
@@ -208,10 +212,10 @@ public function getTag() {
   }
 
   /**
-   * Returns the event on which this rule will trigger.
+   * Returns the array of events on which this rule will trigger.
```
fago: Misses detailed @return documentation. Also, let's better describe the method as "Gets configuration of all events the rule is reacting on." - so it's more clear, that the structure is some configuration structure.

Also, for a lot of code only the event names are from interest (e.g. ReactionRuleStorage), so I think there should be a getter getEventNames() which returns the array of fully qualified event names of the rule.

src/Entity/ReactionRuleConfig.php
```diff
@@ -208,10 +213,30 @@ public function getTag() {
...
 
   /**
-   * Returns the event on which this rule will trigger.
+   * Gets configuration of all events the rule is reacting on.
+   *
+   * @return array
+   *   The events array. The array is numerically indexed and contains arrays
+   *   with the following structure:
+   *     - event_name: string with the event machine name.
+   *     - configuration: an array containing the event configuration.
```
fago: Description sentences should start upper-case

src/Entity/ReactionRuleConfig.php
```diff
@@ -121,11 +121,16 @@ class ReactionRuleConfig extends ConfigEntityBase {
...
-   * @var string
+   * Events array. The array is numerically indexed and contains arrays with the
+   * following structure:
+   *   - event_name: String with the event machine name.
+   *   - configuration: An array containing the event configuration.
+   *
+   * @var array
    */
-  protected $event;
+  protected $events;
```
klausi: We should initialize this as empty array with "[]".

src/Entity/ReactionRuleConfig.php
```diff
@@ -121,11 +121,16 @@ class ReactionRuleConfig extends ConfigEntityBase {
...
 
   /**
-   * The event name this reaction rule is reacting on.
+   * The events this reaction rule is reacting on.
    *
-   * @var string
+   * Events array. The array is numerically indexed and contains arrays with the
+   * following structure:
+   *   - event_name: String with the event machine name.
+   *   - configuration: An array containing the event configuration.
```
klausi, line 124: This is a code smell if we need to document array keys. I think we should use a simple class instead which has this properties + methods documented. We should prevent nested array weirdness when we can.

fago, line 124: Agreed, but can we do that with CMI? I've only seen it save/load arrays.
As alternative, we could use an order array with

> $event_name => $configuration

which Alex tried initially. Problem there was that we need to have dots in event names (as symfony events use dots), but CMI does not allow dots in config keys. Thus, we'd have to massage the event names and somehow escape/replace dots. That makes it more complicated, also when querying for events etc.

klausi, line 124: we can file a follow-up to explore using objects instead of event arrays.

PR #403, opened 2016-02-10
Issue #2409059: Porting Send Mail Functionality (continues PR #390)

emclaughlin1215:
Issue #2409059: Fixing up the tests so that they pass and fixing up the action so it works. Other actions with logging do not have translatable log messages, so I opted to remove that, which made testing easier.

Following up on work done on #390

PR #408, opened 2016-02-18
Issue #2409059: Porting Send Mail Functionality (continues PR #403)

naveenvalecha:
See https://www.drupal.org/node/2409059

Issue #2409059: Fixing up the tests so that they pass and fixing up the action so it works. Other actions with logging do not have translatable log messages, so I opted to remove that, which made testing easier.

Following up on work done on #403 

There are few random failures I'm getting https://travis-ci.org/naveenvalecha/rules/builds/110084721


tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * Data provider for self::testSendMailToRole().
+   */
+  public function providerSendMailToRole() {
+    // Create mock roles here instead of in setUp because this runs before
+    // setUp.
+    foreach (['administrator', 'editor'] as $role_name) {
+      // Mock the role entity.
+      $role_mock = $this->prophesize(RoleInterface::class);
+      $role_mock->id()
```
klausi, line 218: $role_mock is never used. The whole foreach loop seems to be not necessary?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,230 @@
...
+    $this->entityTypeManager->createInstance()
+      ->willReturn($this->userStorage->reveal());
+
+    // Prepare a user entity type instance.
+    $entity_type = reset($this->users)->getEntityType();
+    $this->entityTypeManager->getDefinitions()
+      ->willReturn($entity_type);
+
+    // @todo this is wrong, the logger is no factory.
+    $this->container->set('logger.factory', $logger_factory->reveal());
```
klausi, line 123: The todo comment seems to be outdated? you are passing a factory here, so this looks ok. I think the comment should be removed.

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,230 @@
...
+      $role_mock = $this->prophesize(RoleInterface::class);
+      $role_mock->id()
+        ->willReturn($role_name);
+      $this->roles[] = $role_mock->reveal();
+
+      // Mock the user entity.
+      $user_mock = $this->prophesizeEntity(UserInterface::class);
+      $user_mock->id()->willReturn($uid);
+
+      $user_entity_type = $this->prophesize(EntityTypeInterface::class);
```
klausi, line 90: why do we need this? Does the test work without this user entity type?

tests/src/Integration/Action/SystemMailToUsersOfRoleTest.php
```diff
@@ -0,0 +1,230 @@
...
+    $this->entityTypeManager->getStorage('user')
+      ->willReturn($this->userStorage->reveal());
+
+    $this->entityTypeManager->createInstance()
+      ->willReturn($this->userStorage->reveal());
+
+    // Prepare a user entity type instance.
+    $entity_type = reset($this->users)->getEntityType();
+    $this->entityTypeManager->getDefinitions()
+      ->willReturn($entity_type);
```
klausi, line 120: Same here, I think this is not necessary. Can you check by removing all the entity type related mocking? We should only mock the user objects themselves and the loadByProperties() call to the user storage, which returns the list of user objects. But that is already done in the test method itself.

src/Plugin/RulesAction/SystemMailToUsersOfRole.php
```diff
@@ -0,0 +1,179 @@
...
+   *   Subject of the email.
+   * @param string $body
+   *   Email message text.
+   * @param string $from
+   *   Email from address.
+   */
+  protected function doExecute(array $roles, $subject, $body, $from) {
+    // Get the role ids and remove the empty ones (in case for example the role
+    // has been removed in the meantime).
+    $rids = array_filter(array_map(function ($role) {
```
Eyal-Shalev, line 136: The roles property can be either an array of entities or an array of entity ids.
Currently we are addressing it as both.

I prefer to use it as an array of ids.

naveenvalecha:
Already taken care by another PR so closing this one

PR #409, opened 2016-02-18
Issue #2671056 by fago: Add RulesUiComponentProviderInterface for con…

fago:
…fig entities providing the edited component.


fago:
Merged.

PR #410, opened 2016-02-18
Add Ui for Rules event handlers and per-bundle entities

milkovsky:


milkovsky:
Moved to https://github.com/fago/rules/pull/431 because of strange commits

PR #420, opened 2016-02-27
Issue #2677018 by fago: Condition and Action expression plugins have …

fago:
…unused context methods. Also improve the documenation of those plugins a bit.


fago:
Merged.

PR #424, opened 2016-02-28
Feature/2676996 map definition

fago:


src/Context/ContextHandlerTrait.php
```diff
@@ -30,61 +32,174 @@
...
+
+    if ($plugin instanceof ContextAwarePluginInterface) {
+      // Getting context values may lead to undocumented exceptions if context
+      // is not set right now. So catch those exceptions.
+      // @todo: Remove ones https://www.drupal.org/node/2677162 got fixed.
+      try {
+        $plugin->refineContextDefinitions($selected_data);
+      }
+      catch (ContextException $e) {
+        if (strpos($e->getMessage(), 'context is required') === FALSE) {
```
klausi, line 161: what is going on here? you check for some kind of exception? so you only throw required context exceptions and eat other silently? Can you add a comment?

fago, line 161: There is a comment above it, doesn't that explain it?

fago:
Merged.

PR #427, opened 2016-03-05
Feature/2681733 assertions

fago:


src/Context/ContextAwarePluginInterface.php
```diff
@@ -21,10 +21,32 @@
...
+   *
+   * Note that metadata is only asserted on configuration time. The plugin has
+   * to ensure that the run-time data matches the asserted configuration if it
+   * has been executed successfully.
+   *
+   * @param \Drupal\Core\TypedData\DataDefinitionInterface[] $selected_data
+   *   An array of data definitions for context that is mapped using a data
+   *   selector, keyed by context name.
+   */
+  public function assertMetadata(array $selected_data);
```
klausi, line 63: Not super happy about the name "assertMetadata". What does it mean? Actually it should be "refineMetadata" because it changes data types of context? So what is the difference to refineContextDefinitions() then?

klausi, line 63: Ah, looking further down gives me a clue: this works on the execution metadata state, while refineContextDefinitions works on the context definitions. So this should be "refineExecutionMetadata"?

tests/src/Kernel/Engine/MetadataAssertionTest.php
```diff
@@ -0,0 +1,95 @@
...
+      ->addContextDefinition('node', ContextDefinition::create('entity:node'))
+      ->checkIntegrity();
+    $this->assertEquals(1, iterator_count($violation_list));
+    $this->assertEquals(
+      'Data selector %selector for context %context_name is invalid. @message',
+      $violation_list->get(0)->getMessage()->getUntranslatedString()
+    );
+
+    // Now add the EntityIsOfBundle condition and try again.
+    $rule = $this->expressionManager->createRule();
```
klausi: Why do you have to create a new rule? Just adding the condition to the existing one should be enough?

fago: true, ok.

src/Context/ContextAwarePluginInterface.php
fago, line 63: Yeah, that's the one difference - the other is that this is only applied when the plugin is successfully executed. I think "assert metdata" describes it pretty well as the plugin asserts us that there is some additional metadata available if it has been executed.
But yeah, some more docs to help developers to figure out which one to use might be good.. In d7 it was data_info_alter() and metadata_assertions() and they had different parameters, so they locked more different. Maybe we should highlight the difference is in WHAT gets changed:
- refine context definitions: own context definitions
- assert metadata: selected data definitions

--> I'd suggest we make assertMetadata() return changed data definitions, that way we know what changed also. Thinking of https://www.drupal.org/node/2681955 also, we might want require the changed data definitions to be clones also?

klausi, line 63: Cloning can be a follow-up, I think.

One sentence on each method what is different between them should help.

fago:
and merged!

PR #428, opened 2016-03-06
Add event caching when events are invoked

klausi:
See https://www.drupal.org/node/2513076


klausi:
Looks like this is not the intended approach, closing.

PR #432, opened 2016-03-07
Feature/config changes

fago:


config/schema/rules.schema.yml
```diff
@@ -10,18 +10,18 @@ rules.component.*:
...
-      type: string
-      label: 'Drupal version'
+    tags:
+      type: sequence
+      label: Tags
+      sequence:
+        type: string
+        label: 'Tag'
+    config_version:
+      type: float
```
klausi: float? This is not really a floating number; if we go with "3.0" notation then I think this should be string

PR #443, opened 2016-04-09
Recursion prevention

klausi:
See https://www.drupal.org/node/2280417


klausi:
not working on this anymore.

PR #446, opened 2016-05-04
2654936 exceptions

Cryt1c:
See https://www.drupal.org/node/2654936

I have merged 8.x-3.x into the stalled pull request (https://github.com/fago/rules/pull/369) and fixed some more global LogicExceptions.
There is a problems when I change the InvalidArgumentException in DataFetcherTest.php and DataFetcher.php to the rules namespace. A global \InvalidArgumentException is thrown further down and therefor also expected in the DataFetcherTest.php. Should this \InvalidArgumentException be encapsuled by a rules InvalidArgumentException?


.gitignore
```diff
@@ -0,0 +1 @@
+/.idea
```
klausi: I don't think we should add this file with this pull request. You can have that file locally and ignore itself even.

d7-tests/rules.test
```diff
@@ -191,8 +191,8 @@ class RulesTestCase extends DrupalWebTestCase {
       $action->execute();
       $this->fail("Execution hasn't created an exception.");
     }
-    catch (RulesEvaluationException $e) {
-      $this->pass("RulesEvaluationException was thrown: ". $e);
+    catch (EvaluationException $e) {
+      $this->pass("EvaluationException was thrown: ". $e);
```
klausi: This is the old Drupal 7 file and should NOT be modified. It is only there for reference purposes and to compare.

src/TypedData/DataFetcher.php
```diff
@@ -98,6 +99,7 @@ public function fetchDataBySubPaths(TypedDataInterface $typed_data, array $sub_p
       $current_selector = implode('.', $current_selector);
       throw new \InvalidArgumentException("Unable to apply data selector '$selector' at '$current_selector': " . $e->getMessage());
```
klausi: here is another global \InvalidArgumentException that you forgot? Make sure to grep the code base for all remaining global exceptions

src/TypedData/DataFetcher.php
```diff
@@ -98,6 +99,7 @@ public function fetchDataBySubPaths(TypedDataInterface $typed_data, array $sub_p
...
     }
+
```
klausi: Unrelated change that can be removed?

fago:
see #457

PR #447, opened 2016-05-07
Cache - https://www.drupal.org/node/2513076

fago:
See https://www.drupal.org/node/2513076

### DIFF
--- a/src/Engine/RulesComponentRepositoryInterface.php
+++ b/src/Engine/RulesComponentRepositoryInterface.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rules\Engine\RulesComponentRepositoryInterface.
+ */
+
+namespace Drupal\rules\Engine;
+
+/**
+ * Interface for the component repository.
+ *
+ * The component repository provides the API for fetching executable Rules
+ * components from cache.
+ */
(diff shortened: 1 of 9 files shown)


klausi:
Tests are failing because you need to remove all the file doc blocks. Can you merge with 8.x-3.x?

src/Engine/RulesComponentRepositoryInterface.php
```diff
@@ -0,0 +1,73 @@
...
+   *   - rules_event: The aggregated components of all reaction rules configured
+   *     for an event, identified by the event name; e.g.,
+   *     'rules_entity_presave'.
+   *   Note, that modules may add further resolvers via tagged services. Check
+   *   the rules.services.yml for an example.
+   *
+   * @return \Drupal\rules\Engine\RulesComponent|null
+   *   The component, or NULL if it is not existing.
+   *
+   * @throws \InvalidArgumentException
```
klausi, line 45: let's stop using global execptions and create one under the rules namespace, see also https://www.drupal.org/node/2654936

fago, line 45: Agreed - but that exception is introduced already by PR-369 so it does not make sense much to do it again here + just cause conflicts? Let's better re-roll whatever PR gets merged once one made it.

PR #450, opened 2016-05-12
Issue #2723259 by yanniboi: Allow data collection input for multiple context fields

yanniboi:
This issue is mirrored on drupal.org by https://www.drupal.org/node/2723259.

Added exceptions allowing single value fields to be passed to multiple context fields.

### DIFF
--- a/src/Context/ContextHandlerIntegrityTrait.php
+++ b/src/Context/ContextHandlerIntegrityTrait.php
@@ -134,7 +134,10 @@ trait ContextHandlerIntegrityTrait {
     $target_type = $context_definition->getDataDefinition()->getDataType();
 
     // Special case any and entity target types for now.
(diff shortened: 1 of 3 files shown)


yanniboi:
So far I have added an exception that allows submitting a single value field to a multiple context field.

Also when a single value is passed to an action or condition, expecting an array, it gets wrapped in an array.

yanniboi:
I've tested my changes by passing '@user.current_user_context:current_user.mail.value' to the SystemSendEmail 'to' field which is multiple cardinality.

Works for me, but needs testing, and I could do with someone confirming that I'm on the right track or letting me know if I am barking up the wrong tree...

yanniboi:
OK, so it looks like my exception for lists is breaking IntegrityCheckTest::testListTypeViolation() which makes sense.

I think what I've dont so far is enough of a proof of concept, but I'm going to leave it here for now, as I don't even know yet if I am going about this the right way. Feel free to comment.

PR #451, opened 2016-05-13
Issue #2724595 by yanniboi: Updated rules send send email action schema.

yanniboi:
Just a small one.

Updated the schema of rules_send_email context value to reflect 'to' being multiple cardinality.

d.o: https://www.drupal.org/node/2724595


klausi:
thanks, tests are missing, continued in #456

PR #453, opened 2016-05-18
email body as text area

AntonMartyniuk:
Make text body field as text area

### DIFF
--- a/src/Context/ContextDefinition.php
+++ b/src/Context/ContextDefinition.php
@@ -28,6 +28,13 @@ class ContextDefinition extends ContextDefinitionCore implements ContextDefiniti
   ];
 
   /**
+   * Type of form element to be used.
+   *
+   * @var string
+   */
+  protected $formElement = 'textfield';
+
+  /**
    * Whether the context value is allowed to be NULL or not.
    *
    * @var bool
--- a/src/Context/ContextDefinition.php
+++ b/src/Context/ContextDefinition.php
@@ -21,6 +21,7 @@ class ContextDefinition extends ContextDefinitionCore implements ContextDefiniti
     'multiple' => 'isMultiple',
     'required' => 'isRequired',
     'default_value' => 'defaultValue',
+    'form_element' => 'formElement',
     'constraints' => 'constraints',
     'allow_null' => 'allowNull',
     'assignment_restriction' => 'assignmentRestriction',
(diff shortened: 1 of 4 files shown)


src/Context/ContextDefinition.php
```diff
@@ -21,6 +21,7 @@ class ContextDefinition extends ContextDefinitionCore implements ContextDefiniti
     'multiple' => 'isMultiple',
     'required' => 'isRequired',
     'default_value' => 'defaultValue',
+    'form_element' => 'formElement',
```
yanniboi, line 24: You haven't added 'formElement'.

AntonMartyniuk, line 24: 'formElement' is at src/Plugin/RulesAction/SystemSendEmail.php

yanniboi, line 24: Not that I can see. You need something like:

```
  /**
   * Type of form element to be used.
   *
   * @var string
   */
  protected $formElement = 'textfield';
```

ContextDefinition::toArray() is trying to set ContextDefinition::$formElement and it doesn't exist yet.

yanniboi:
Almost there! I tested with simplytest.me and found that the email message still wasn't a textarea. After looking it turns out that the problem was that $formElement is a protected variable so ContextFormTrait cant use it directly.

I have create a fix for this here: https://github.com/AntonMartyniuk/rules/pull/2

src/Context/ContextDefinitionInterface.php
```diff
@@ -61,6 +61,24 @@ public function getAssignmentRestriction();
   public function setAssignmentRestriction($restriction);
 
   /**
+   * Determines form element for the context type.
+   *
+   * @return bool
+   *   Type of form element to be used.
```
klausi, line 68: return type is wrong, this should be "string", right?

src/Context/ContextDefinitionInterface.php
```diff
@@ -61,6 +61,24 @@ public function getAssignmentRestriction();
...
+   */
+  public function getFormElement();
+
+  /**
+   * Sets the form element for context.
+   *
+   * @param string $form_element
+   *   Type of form element to be used.
```
klausi, line 76: Let's add an example, '... , for example "textarea"."'

src/Context/ContextDefinitionInterface.php
klausi, line 68: Let's add an example, '... , for example "textarea".'

PR #467, opened 2016-08-10
Added action plugin: Fetch entities by view

pmelab:
I've added a plugin that enables rules to fetch lists of entities from view. Similar to the action available in [Rules bonus pack](https://www.drupal.org/project/rb), but with typed contexts.

There is a views display plugin to be used for views that act as rules data sources. A plugin deriver will provide a rules action for every views display using this plugin, and add views contexts based on the views  argument definitions.

I'm aware that there is no test coverage and probably some bugs. I would like to discuss feasibility of this approach before going any further.

### DIFF
--- a/src/Plugin/RulesAction/EntityFetchByView.php
+++ b/src/Plugin/RulesAction/EntityFetchByView.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Drupal\rules\Plugin\RulesAction;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Entity\Plugin\DataType\EntityAdapter;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\rules\Core\RulesActionBase;
+use Drupal\views\Views;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a generic 'Fetch entities by view' action.
+ *
+ * @RulesAction(
+ *   id = "rules_entity_fetch_by_view",
+ *   deriver = "Drupal\rules\Plugin\RulesAction\EntityFetchByViewDeriver",
+ *   category = @Translation("Entity")
+ * )
+ */
+class EntityFetchByView extends RulesActionBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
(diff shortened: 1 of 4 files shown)


src/Plugin/RulesAction/EntityFetchByView.php
```diff
@@ -0,0 +1,106 @@
...
+
+    $view_id = $this->pluginDefinition['view_id'];
+    $display_id = $this->pluginDefinition['display_id'];
+
+    // Fetch the list of available contexts.
+    $contexts = $this->getContexts();
+
+    // Pull values out of contexts.
+    $contexts = array_map(function ($context) {
+      return $context->getContextData()->getValue();
```
fago, line 78: This can be nice done by implementing doExecute() instead.

src/Plugin/RulesAction/EntityFetchByView.php
```diff
@@ -0,0 +1,106 @@
...
+    }, $contexts);
+
+    // Convert entities into entity ids.
+    $contexts = array_map(function ($context) {
+      return $context instanceof EntityInterface ? $context->id() : $context;
```
fago, line 83: Interesting. Mabye we should add some support to allow ids to be required only.

src/Plugin/RulesAction/EntityFetchByViewDeriver.php
```diff
@@ -0,0 +1,121 @@
...
+      if ($base_table = $entity_type->getBaseTable()) {
+        $entity_types[$base_table] = $entity_type;
+      }
+
+      if ($data_table = $entity_type->getDataTable()) {
+        $entity_types[$data_table] = $entity_type;
+      }
+    }
+
+    foreach (Views::getApplicableViews('rules') as $data) {
```
fago, line 83: ouch, is there no method on the storage we can use?

src/Plugin/views/display/Rules.php
```diff
@@ -0,0 +1,143 @@
...
+ *   title = @Translation("Rules"),
+ *   admin = @Translation("Rules entity source"),
+ *   help = @Translation("Provide views results to rules workflows."),
+ *   theme = "views_view",
+ *   register_theme = FALSE,
+ *   uses_menu_links = FALSE,
+ *   rules = TRUE
+ * )
+ */
+class Rules extends DisplayPluginBase {
```
fago, line 20: Let's call this "Rules context provider" ? Theoretically this could be used by other modules also.

fago: That said, maybe does Ctools already have something in that direction we could use instead?

fago (changes requested):
Thanks for the PR - this looks great already. I started with a review!

PR #469, opened 2016-08-20
Update docs link in readme

tormi:


fago:
The actual correct link is docs.d8rules.org

PR #452, opened 2016-05-13
Issue #2648334 - Drag and drop reordering for actions/conditions

yanniboi:
This is mirrored on d.o here: https://www.drupal.org/node/2648334

PR #475, opened 2016-10-04
Issue #2799361 - Port condition expression action links to the UI. 

dsasser:
Adds the ability to group conditions within AND and OR condition sets in the UI. I used the patch from #452 to provide drag and drop functionality to the form and modified from there. See the d.o issue at: https://www.drupal.org/node/2799361

### DIFF
--- a/src/Form/Expression/ActionContainerForm.php
+++ b/src/Form/Expression/ActionContainerForm.php
@@ -39,31 +39,75 @@ class ActionContainerForm implements ExpressionFormInterface {
     ];
 
     $form['action_table']['table'] = [
-      '#theme' => 'table',
-      '#caption' => $this->t('Actions'),
-      '#header' => [$this->t('Elements'), $this->t('Operations')],
-      '#empty' => t('None'),
+      '#type' => 'table',
+      '#header' => [
+        $this->t('Elements'),
+        $this->t('Weight'),
+        [
+          'data' => $this->t('Operations'),
+          'colspan' => 3,
+        ],
+      ],
+      '#attributes' => [
+        'id' => 'rules_actions_table',
+      ],
+      '#tabledrag' => [
+        [
+          'action' => 'order',
+          'relationship' => 'sibling',
+          'group' => 'action-weight',
+        ],
+      ],
     ];
 
+    $form['action_table']['table']['#empty'] = $this->t('None');
+
+    // Get hold of actions.
+    // @todo See if we can add getExpressions method of ExpressionContainerBase.
+    $actions = [];
     foreach ($this->actionSet as $action) {
-      $form['action_table']['table']['#rows'][] = [
-        'element' => $action->getLabel(),
-        'operations' => [
-          'data' => [
-            '#type' => 'dropbutton',
-            '#links' => [
-              'edit' => [
-                'title' => $this->t('Edit'),
-                'url' => $this->getRulesUiHandler()->getUrlFromRoute('expression.edit', [
-                  'uuid' => $action->getUuid(),
-                ]),
-              ],
-              'delete' => [
-                'title' => $this->t('Delete'),
-                'url' => $this->getRulesUiHandler()->getUrlFromRoute('expression.delete', [
-                  'uuid' => $action->getUuid(),
-                ]),
-              ],
+      $actions[] = $action;
+    }
+
+    // Sort actions by weight.
+    @uasort($actions, [$this->actionSet, 'expressionSortHelper']);
+
+    foreach ($actions as $action) {
+      /* @var $action \Drupal\rules\Engine\ExpressionInterface */
+      $uuid = $action->getUuid();
+      $row = &$form['action_table']['table'][$uuid];
+
+      // TableDrag: Mark the table row as draggable.
+      $row['#attributes']['class'][] = 'draggable';
+
+      // TableDrag: Sort the table row according to its existing weight.
+      $row['#weight'] = $action->getWeight();
+      $row['title'] = ['#markup' => $action->getLabel()];
+
+      $row['weight'] = [
+        '#type' => 'weight',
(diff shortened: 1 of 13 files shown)


src/Form/Expression/ActionContainerForm.php
```diff
@@ -34,36 +34,80 @@ public function __construct(ActionExpressionContainerInterface $action_set) {
...
+        '#delta' => 50,
+        '#default_value' => $action->getWeight(),
+        '#attributes' => ['class' => ['action-weight']],
+      ];
+
+      // Operations (dropbutton) column.
+      $rules_ui_handler = $this->getRulesUiHandler();
+      $row['operations'] = [
+        'data' => [
+          '#type' => 'dropbutton',
```
kevin-dutra: Technically not part of the new changes, but there's a dedicated `Operations` render element that extends from `Dropbutton`.

src/Form/Expression/ConditionContainerForm.php
```diff
@@ -24,65 +25,356 @@ class ConditionContainerForm implements ExpressionFormInterface {
...
+   *
+   * @param array $values
+   *   The array of form values.
+   *
+   * @return array|mixed
+   *   The derived condition expression elements.
+   */
+  private function mirrorElements($values) {
+    $elements = NULL;
+    $expressionManager = \Drupal::service('plugin.manager.rules_expression');
```
kevin-dutra: Ideally this service would be introduced via dependency injection, rather than doing the global call. As an example, see `ContentEntityForm`.

src/Plugin/RulesExpression/ActionSet.php
```diff
@@ -27,7 +27,17 @@ protected function allowsMetadataAssertions() {
    * {@inheritdoc}
    */
   public function executeWithState(ExecutionStateInterface $state) {
+    // Get hold of actions.
+    // @todo See if we can add getExpressions method of ExpressionContainerBase.
+    $actions = [];
```
kevin-dutra, line 32: Is it really necessary to put in an extra loop just for the copying to a local variable? Can't you just
`$actions = $this->actions;` or even just sort the class property directly? (Ditto for the other applicable classes.)

src/Form/Expression/ConditionContainerForm.php
```diff
@@ -24,65 +25,356 @@ class ConditionContainerForm implements ExpressionFormInterface {
...
+      $form['init_help'] = array(
+        '#type' => 'container',
+        '#id' => 'rules-plugin-add-help',
+        'content' => array(
+          '#markup' => t('You are about to add a new @plugin to the @config-plugin %label. Use indentation to make conditions a part of this logic group. See <a href="@url">the online documentation</a> for more information on condition sets.',
+            array(
+              '@plugin' => $this->conditionContainer->getLabel(),
+              '@config-plugin' => $config->bundle(),
+              '%label' => $config->label(),
+              '@url' => 'http://drupal.org/node/1300034',
```
kevin-dutra: There is a newer placeholder type that should be used here, I think: https://www.drupal.org/node/2571689

kevin-dutra (changes requested):
A few minor things that I can see.

src/Form/Expression/ConditionContainerForm.php
dsasser: I agree that dependency injection would be ideal. The reason its not implemented with DI is because the form is being called in a non-standard way (not from the container). Its getting a single argument passed into the constructor, and so we would have to go further up into the api to affect that. I wanted to stay away from making api changes as much as possible.

src/Form/Expression/ConditionContainerForm.php
dsasser: Good to know about the newer placeholder. I assume we are talking about the @url element? Since its not user provided input I wouldn't normally worry about XSS escaping for a hard coded url, but I can change it to use the new placeholder quick enough that its a moot point.

kevin-dutra: Yep, `@url`.

src/Form/Expression/ConditionContainerForm.php
kevin-dutra: Ah, okay. Maybe abstract it out into a separate method, e.g.

```
$expressionManager = $this->getExpressionManger();
...
function getExpressionManger() {
  return \Drupal::service('plugin.manager.rules_expression');
}
```

Mainly I'm just trying to keep options open for unit testing.

src/Form/Expression/ActionContainerForm.php
dsasser: The dropdown was part of drag/drop PR #452 that this PR is using. But knowing that the `operations` element adds a theme suggestion is a good reason to change it.

src/Plugin/RulesExpression/ActionSet.php
dsasser: Actually I think its necessary because $this->actions is really an iterable object and not an array. This is part of the code coming from the included PR #452.

kevin-dutra: Hmm...looks like an array to me:

```
  /**
   * List of actions that will be executed.
   *
   * @var \Drupal\rules\Engine\ActionExpressionInterface[]
   */
  protected $actions = [];

  ...

  public function __construct(array $configuration, $plugin_id, $plugin_definition, ExpressionManagerInterface $expression_manager) {

    ...

    foreach ($configuration['actions'] as $action_config) {
      $action = $expression_manager->createInstance($action_config['id'], $action_config);
      $this->actions[] = $action;
    }
  }
```

But at any rate, if it's part of the other PR, then I won't split hairs here.

kevin-dutra (approved):
Cool beans.

PR #476, opened 2016-10-10
#2815967 - Lookup user roles before trying to execute UserRoleAdd act…

dsasser:
The AddUserRole action is broken. This should solve it until typed data widgets are updated.


dsasser:
See also https://www.drupal.org/node/2816157#comment-11717929

PR #477, opened 2016-10-10
#2816157 - Lookup user roles by id and label when trying to evaluate …

dsasser:
…UserHasRole condition.


dsasser:
See https://www.drupal.org/node/2816157#comment-11717929 for more information.

PR #481, opened 2017-04-18
Issue #2849779: Drupal 8.3 Critical - Implement missing PluginDefinitionInterface methods in RulesUiDefinition.

esolitos:
Rules on Drupal 8.3.0 breaks completely, this PR comes directly from d.o.
Ref. https://www.drupal.org/node/2849779

### DIFF
--- a/src/Ui/RulesUiDefinition.php
+++ b/src/Ui/RulesUiDefinition.php
@@ -111,6 +111,20 @@ class RulesUiDefinition implements PluginDefinitionInterface {
   /**
    * {@inheritdoc}
    */
+  public function id() {
+    return $this->id;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getProvider() {
+    return $this->provider;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function setClass($class) {
     $this->class = $class;
     return $this;


xuxucode:
The tests in **Integration** folder seem outdated on 8.3.x

jonathan1055:
Thanks for the PR esolitos.
Yes, the tests at Core 8.1 and 8.2 seem OK but with 8.3 we get 
`PHP Fatal error:  Class 'Drupal\Tests\rules\Integration\RulesIntegrationTestBase' not found in /home/travis/build/fago/rules/tests/src/Integration/Action/BanIPTest.php on line 13`

jonathan1055:
I get exactly the same error running the integration tests locally, using sourcecode from d.o.

esolitos:
I admit that I have not run the tests for the patch.
However I think that it's important to have rules not to break the whole drupal site over fixing the integration tests, so I suggest first just "hot-patching" the `RulesUiDefinition` class, and taking care of the tests in a second moment.
ATM any site upgrading to 8.3 with rules, will simply break. :)

jonathan1055:
I agree totally. We know that this patch will fix Rules to allow it to be installed at 8.3. The test can be fixed later - although it does need to be fixed soon as it nullifies all other rules patch testing.

jhedstrom:
I filed https://www.drupal.org/node/2870797 as these tests fail with or without this patch.

PR #482, opened 2017-04-18
Move tests to a supported namespace.

jhedstrom:
This gets the integration tests passing again by moving them into a subdirectory of the unit tests. Apparently core 8.3.x and beyond only supports the 4 test namespaces of `Drupal\Tests\*\Unit`, `Drupal\Tests\*\Kernel`, `Drupal\Tests\*\Functional`, and `Drupal\Tests\*\FunctionalJavascript`.

See https://www.drupal.org/node/2870797.


jhedstrom:
Note the remaining failures are addressed by #481.

jhedstrom:
I've merged the fix from #481 to see if tests pass.

jonathan1055:
That's great. I have also cofirmed this manually, and added my results to https://www.drupal.org/node/2870797#comment-12045359

What do we have to do to get a Rules maintainer to commit these two patches? I know there has been a funding problem to pay for their time.

PR #484, opened 2017-04-27
Issue #2824348

jonathangreen:
Updating the patch from Drupal.org and making pull request. Issue on Drupal.org:
https://www.drupal.org/node/2824348


jonathan1055:
Thanks @jonathangreen for doing this. Now atleast we can see the tests running and investigate the failure.

jonathangreen:
@fago I updated the patch from d.o [here](https://www.drupal.org/node/2824348#comment-12047585) and updated the prophecy so that it reflects the new call to `getContextDefinition` in the patch. Should be passing all tests now.

PR #486, opened 2017-08-22
Issue #2903855 Allow travisCI builds not to fail for just coding standards messages

jonathan1055:
This is the pull request for Rules issue https://www.drupal.org/node/2903855
"Allow travisCI builds not to fail for just coding standards messages"
Two runtime parameters are added to the call to phpcs so that coding standards messages are displayed but do not influence the pass/fail of the tests.


jonathan1055:
The builds at core 8.2 and 8.3 now pass, because the tests pass (and coding standards exit code is ignored). The builds at core 8.4 fail, because the tests actually fail. This is as expected.

PR #488, opened 2017-09-26
Rules issue #2903855 fix coding standards 

jonathan1055:
This was the original pull request for Drupal Rules issue https://www.drupal.org/node/2903855

It has been discarded.

PR #489, opened 2017-09-28
#2903855 Fix coding standards

jonathan1055:
This is a pull request for Drupal Rules issue https://www.drupal.org/node/2903855
In addition to fixing the coding standards violations in the various code files, the other changes necessary to achive this are:
- Run tests at Core 8.5 and drop testing 8.2
- Add commands to list the sniffs and run a phpcs summary
- Set the required Coder version to 8.2.12. It will be up to us to increase this when 8.2.13 is available.
- README.md is clean when uing the latest dev of Coder, but the lineLength sniff is excluded until we are using Coder 8.2.13
- Travis build does not run ESlint so I did not notice that there are 4 more warnings on the drupal.org testbot output for js/autocomplete.js. These will also be fixed next.

With the type-hint fix (commit https://github.com/fago/rules/pull/488/commits/306535b324953a47d3ca7a7857f0c55d75ba302d) one suggestion to add array caused runtime errors. This was fixed by removing [ ] from the @ param tag - see later commit https://github.com/fago/rules/pull/488/commits/d550b9c65bd0be9be0a49712de826064a42ed890


src/Core/ConditionManager.php
```diff
@@ -13,16 +13,6 @@ class ConditionManager extends CoreConditionManager {
 
   /**
    * {@inheritdoc}
-   *
-   * @return \Drupal\rules\Core\RulesConditionInterface|\Drupal\Core\Condition\ConditionInterface
```
fago, line 17: This should not be removed. It's an helpful type-hint!

fago (changes requested):

src/Core/ConditionManager.php
jonathan1055: OK, I was wondering about that. I guess the coder rule is intended so that you do not duplicate info that is also in the ancestor function which the inheritdoc is pointing to. So either we write out the full comment block without having {inheritdoc} or I can exclude that sniff from this specific file. I traced the inheritance from ConditionManager -> DefaultPluginManager -> PluginManagerBase where getDiscovery() just has 
```
/**
 * Gets the plugin discovery.
 *
 * @return \Drupal\Component\Plugin\Discovery\DiscoveryInterface
 */
```
Let me know your preference and I'll make the change.

jonathan1055: Ah, I was looking at the following function getDiscovery(). Sorry. The removal of the return is actually misleading, because the patch has split two functions. What is actually happening is the entire `createInstance()` function is being deleted as a "useless override".  Did you mean that this function should not be removed?

fago:
This got already merged.

jonathan1055:
Yes that's great. Thanks.

PR #491, opened 2017-09-29
Issue #2912782 : Improve test @group annotations

dman-coders:
<h3>Approach: 
</h3>Generally, 
* anything in the functional tests (BrowserTestBase) should probably be called RulesUi.
* anything lower-case 'rules' needs to become title case 'Rules'.
* Where sensible, additional sub labels may make sense. The each of the RulesAction and RulesCondition and RulesEvent tests (now singular) are already in convenient directories, so keep that pattern.

<h3>Gotcha:
</h3>
RulesLoggerChannelTest was in @group 'Logger' - it doesn't seem conventional to be adding our tests under some other modules namespace, so I set it back to 'Rules'.
There was a stray @group inside one of the test objects functions AnnotationProcessingTest->testCheckConfiguration(), serving no purpose. Removed it.

<h3>Before patch:</h3>
<code>
../../vendor/bin/phpunit --list-groups ../modules/contrib/rules/tests/src/

Available test group(s):
 - Logger
 - Rules
 - rules
 - rules_action
 - rules_actions
 - rules_conditions
 - rules_events
 - rules_ui
</code>

<h3>After patch:</h3>
<code>
../../vendor/bin/phpunit --list-groups ../modules/contrib/rules/tests/src/

Available test group(s):
 - Rules
 - RulesAction
 - RulesCondition
 - RulesEvent
 - RulesUi
</code>


jonathan1055:
Thanks @dman-coders this is a nice tidy-up.
Just for reference the d.o. issue is https://www.drupal.org/node/2912782

PR #493, opened 2017-10-15
#2913228 Remove require-dev and adjust .travis.yml

jonathan1055:
Drupal issue https://www.drupal.org/node/2913228
This PR removes the recent "require-dev" additions in composer.json and adjusts .travis.yml to use the phpcs which is already installed as part of the 'composer install'. This fix is to allow testing on D.O. jenkins builds to run again. Then we can look at introduing require-dev requirements later, and the patches will be tested on d.o.


jonathan1055:
This PR is no longer required now that tests on d.o. have suddenly started working again without any changes to Rules code.

PR #497, opened 2017-11-27
Issue #2922757 Add @group legacy to 20 test files to avoid deprecation failures

jonathan1055:
Add @group legacy to 4 functional and 16 kernel test classes to avoid failure for deprecated RouteEnhancerInterface

The d.o. issue is https://www.drupal.org/project/rules/issues/2922757


jonathan1055:
Fago committed this change via patch 30 Nov.

PR #498, opened 2017-11-28
BanIP action causes PHP Exception, issue 2922804

jonathan1055:
D.O. issue https://www.drupal.org/project/rules/issues/2922804


jonathan1055:
Fago committed this on 14 Dec 2017

